PR #13, opened 2014-05-19
Chinese lang support in 

braveh4rt:
Hi avleen ,

I had found the solution to my problem with chinese lang email come to exim.

you can check my change in the exim configuration

Thanks

Mulianto


rimas-kudelis:
Since I haven't got any response to my questions from the original author for more than six months and because I don't think this is the right solution to use, I'm just going to close this.

PR #24, opened 2014-09-14
Spamassassin tuning, ignore big files and add feature to reject mails wi...

Udera:
...th high spam-score directly.

In some old configs, I found some improvements concerning the handling with spam mail:
- To save resources, only mails larger than 100 KB are scanned because typical spam is only about a few KB (per default on).
- The second option is to reject mails with spam scores higher than 15 directly. The interest is to avoid the handling of mails with a high spamassassin score. This concerns especially addresses which are forwarded to other  servers (e.g. googlemail). These forwarded messages are often rejected by the other server, then our server creates bounce-massages (backscatter) and can obtain a bad reputation. It is off by default. You have to verify first, that spamassassin is properly working, 15 is a more or less random value.

### DIFF
--- a/docs/vexim-acl-check-content.conf
+++ b/docs/vexim-acl-check-content.conf
@@ -25,7 +25,7 @@
   
   # Mails larger than 100 KB are accepted without spam-checking to save resources.
   # Typical Spam has only a few KB in size.
-  accept  condition = ${if >={$message_size}{100k}{yes}{no}}
+  #accept  condition = ${if >={$message_size}{100k}{yes}{no}}
 
   # Reject mails with more than 15 spam-points. This is a system-wide setting and does not respect
   # individual user settings!
--- a/docs/vexim-acl-check-content.conf
+++ b/docs/vexim-acl-check-content.conf
@@ -22,6 +22,16 @@
   # WARNING: this is an example !
   # deny  message = This message matches a blacklisted regular expression ($regex_match_string)
   #      regex = [Vv] *[Ii] *[Aa] *[Gg] *[Rr] *[Aa]
+  
+  # Mails larger than 100 KB are accepted without spam-checking to save resources.
+  # Typical Spam has only a few KB in size.
+  accept  condition = ${if >={$message_size}{100k}{yes}{no}}
+
+  # Reject mails with more than 15 spam-points. This is a system-wide setting and does not respect
+  # individual user settings!
+  #deny    message = This message has been rejected due to spam.
+  #        spam = maildeliver:true
+  #        condition = ${if >{$spam_score_int}{150}{true}{false}}
 
   # Always add X-Spam-Score and X-Spam-Report headers, using SA system-wide settings
   # (user "nobody"), no matter if over threshold or not.


rimas-kudelis:
I just looked at the patch again. I don't mind taking it in, if you leave both of these options commented out by default.

The reason is, I don't think that scanning big e-mails for spam affects typical mail server performance in any significant way (does it?), so I'd rather leave this feature as a draft for those admins, who are actually worried about it.

Udera:
Today it's probably less a performance issue. But for some admins, this could be an interesting feature. I'm ok to comment it out by default.

rimas-kudelis:
The comments could use some improvement, I guess (particularly, I would expect something like "Uncomment this to..." for cases with non-default settings commented out, but overall, that's cool.

I will be absent from 9th to 16th of May, so accepting this patch as it is now. Thanks!

rimas-kudelis:
It was my request to disable this. 
If SA is indeed being killed, then maybe you're right, I didn't know about that. However, 100KB seems a little too little a limit to me. I mean, I don't think it would take 5 seconds or more to scan 100KB for spam.

soujak:
I missed the previous discussion. Sorry about that.

I will check the default Exim timeout and the default SA and report in the next days.

On old hardware and or high volume servers I personally experimented the need to lower this limit in order to limit performance degradation of SA. On the other hand, one may also think about the opposite argument of @rimas-kudelis : how many spam messager are bigger than, say, 250 KB? (This looks as a reasonable size.)

rimas-kudelis:
Probably not many.
But I'm worried that should disabling spam checks for big messages become popular, spammers might begin sending out bigger e-mails, just to avoid being scanned.

soujak:
On the **default limits** of surrounding softwares. SpamAssassin's `spamc` has 500 KB as default limit (cf. [SA v3.4 documentation about it](https://spamassassin.apache.org/full/3.4.x/doc/spamc.html)). Exim recent versions have a default timeout of 2 minutes (cf. [Exim4 documentation](http://www.exim.org/exim-html-current/doc/html/spec_html/ch-content_scanning_at_acl_time.html)) whilst older ones used to have 5 minutes (cf. [O'Reilly Commons page](http://commons.oreilly.com/wiki/index.php/SpamAssassin/Integrating_SpamAssassin_with_Exim#Setting_a_timeout_on_spamc)). (I previously wrote 5 seconds, but I meant to say minutes! :smiley:  Sorry.) For this reason, SA-Exim, being indeed a bit outdated, has an aggressive default policy designed for older Exim installation -- after a timeout of 4 minutes it gives a temporary SMTP rejection.

On the **risk of facilitating spam** traffic with the the limit. An average SPAM message is less than 1 KB, that being a quite conservative estimate. Bandwidth usage comes with a price for spammers as well, so I do not think they can afford the cost (time or money) to increase their cost of two orders of magnitude just to cheat SpamAssassin (send 1/100 the volume of mail, or pay 100 times the connection services).
By the way, another argument is more simplistically given by the fact that every mail server in the Internet must have a resource limit of some kind, timeout or maximum message size.

Udera:
I have some Spam mails of 300-400 kB in size, the limit should probably 
a bit higher than the 250 kB recommended here: 
https://wiki.apache.org/spamassassin/OutOfMemoryProblems#Very_large_mails. 
The limit can always be raised or removed once there are larger spam 
mails.

I can understand the other strategy of no limits at all. Perhaps we 
should check the performance of spamassassin if you send a mail with a 
large attachment.

rimas-kudelis:
Allright, I'm convinced we should keep a limit. :)

PR #25, opened 2014-10-24
Pipe cor

Udera:
In the dovecot-documentation, there are exim-variables used in the pipe command (http://wiki2.dovecot.org/LDA/Exim). To make them work, we need to expand the expression obtained from the database. One could directly write the users-address but it won't work for the $sender_address:
/usr/local/libexec/dovecot/dovecot-lda -d $local_part@$domain -f $sender_address

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -586,12 +586,12 @@ virtual_forward:
 virtual_domains:
   driver = redirect
   allow_fail
-  data = ${lookup mysql{select smtp from users,domains \
+  data = ${expand:${lookup mysql{select smtp from users,domains \
   		where localpart = '${quote_mysql:$local_part}' \
 		and domain = '${quote_mysql:$domain}' \
 		and domains.enabled = '1' \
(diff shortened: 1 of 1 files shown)


Udera:
I could make a separate route for all users that want to pass mails directly to dovecot (I select them by a modified SQL-query, on_piped=1). This is nice in my case, but if someone wants to pipe his mails to a different program?

rimas-kudelis:
OK, I've been deferring my decision, but: I really don't like the idea of giving all users the right to run arbitrary commands on the mail server, using mail daemon's permissions.
It would probably be acceptable if the users, which are allowed to run commands, were appropriately flagged in the database (like in your last example, `on_piped`), but I'm not too enthusiastic about that idea either, because I'm simply not sure how common such cases are.
I'm closing this PR, but feel free to make another one, with a flag in the database and a cleaner commit history. If you do, please reference this PR in it, so that people can easily find this discussion.
Anyway, thanks for contributing!

PR #29, opened 2015-01-03
Add submission port 587 and SSL for incoming SMTP connections.

Udera:
Today, the use of the submission port or SSL port can be activated right away.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -55,9 +55,10 @@ MAILMAN_GROUP=mail
 
 # SMTP port 25 is activated by default. This port is blocked in some networks
 # and it is recommended to use the submission port 587 for an authenticated
-# SMTP-connection or use SSL.
-daemon_smtp_ports = 25 : 465 : 587
-tls_on_connect_ports = 465
+# SMTP-connection. SSL port 465 is deprecated, in order to use it, add port 465
+# to the daemon_smtp_ports and uncomment the line below (tls_on_connect ports)
+daemon_smtp_ports = 25 : 587
+#tls_on_connect_ports = 465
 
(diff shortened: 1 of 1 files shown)


Z3po:
Using the SSL-only Port is deprecated. No need to implement stuff that shouldn't be used anymore. Implementing port 587 seems ok to me.

rimas-kudelis:
@Udera: what happens with `tls_on_connect_ports` if SSL is not configured? Will Exim still work?

@Z3po: I'm not sure if having 465 enabled in default config would hurt. For what it's worth, I'm using exact same lines myself.

Z3po:
Well it wouldn't hurt but according the port *is* deprecated the port number itself has been re-assigned by IANA to some other protocoll I don't think it makes much sence to integrate that in a newer version of Vexim. I totally agree to enable port 587 anyhow.

rimas-kudelis:
OK, I'm convinced. @Udera, would you mind removing port 465 from the default config, or at least commenting it out by default?

Udera:
@rimas-kudelis I updated the pull request.

docs/configure
```diff
@@ -53,6 +53,13 @@ MAILMAN_GROUP=mail
 
 # primary_hostname =
 
+# SMTP port 25 is activated by default. This port is blocked in some networks
+# and it is recommended to use the submission port 587 for an authenticated
+# SMTP-connection. SSL port 465 is deprecated, in order to use it, add port 465
+# to the daemon_smtp_ports and uncomment the line below (tls_on_connect ports)
```
rimas-kudelis: no need for the first "the" here. :)

PR #32, opened 2015-02-27
adding debianic conf.d files

Z3po:
I've been going through the configuration file and are now adding the conf.d files I'm using in my own Setup.


docs/debian-conf.d/auth/01_exim4-config_vexim
```diff
@@ -0,0 +1,24 @@
+
+### 01_exim4-config_vexim
+#########################
+
+plain_login:
```
rimas-kudelis: Can you rename these to `plain_vexim` and `login_vexim`?
I'm guessing that these names must be unique, is that correct? In that case, having `vexim` in them seems a natural choice.

docs/debian-conf.d/auth/01_exim4-config_vexim
```diff
@@ -0,0 +1,24 @@
...
+        driver = plaintext
+        public_name = PLAIN
+        server_condition = "${if crypteq{$3}{${lookup mysql{ \
+                          SELECT crypt FROM users \
+                          WHERE username = '${quote_mysql:$2}' \
+                          }}}{yes}{no}}"
+        server_set_id = $2
+        server_advertise_condition = ${if eq{$tls_cipher}{}{}{*}}
```
rimas-kudelis: I guess these login drivers should basically match those specified in docs/configure. That means that the advertise_condition should also match. See https://github.com/avleen/vexim2/blob/master/docs/configure#L932 and below.

docs/debian-conf.d/router/250_vexim_virtual_domains
```diff
@@ -65,8 +65,9 @@ virtual_domains:
                                and users.on_spamassassin = 0 \
                                and users.domain_id=domains.domain_id}{$value}fail}}} \
                            } {X-Spam-Score:X-Spam-Report} }
-  local_part_suffix = -*
-  local_part_suffix_optional
+  # optional!
+  #local_part_suffix = -*
+  #local_part_suffix_optional
```
rimas-kudelis: I think these should be enabled by default, cause they are enabled in docs/configure.
On the other hand, IF these lines aren't specific to vexim routers, they could probably be removed altogether. However, it seems they are specific to our routers (aren't they?), so they should probably stay.

docs/debian-conf.d/transport/30_vexim_virtual_delivery
```diff
@@ -1,6 +1,6 @@
 
 ### transport/30_vexim_virtual_delivery
-#################################
+#######################################
 
 virtual_delivery:
```
rimas-kudelis: Can you please rename all these transports to have 'vexim_' prefix in their name?

rimas-kudelis:
Other than my comments, I'm glad that someone other than me is using vexim under Debian, and is actively contributing to debian conf.d files. Thank you! :)

docs/debian-conf.d/transport/30_vexim_virtual_delivery
Z3po: Well at least in that case I'm equal to the default "configure" file :)
I'm keeping that "as is" because we're having "virtual" in the name.
There's currently no real "standard" about how things should be named...

Z3po:
I've added another commit with the suggested changes.

docs/debian-conf.d/transport/30_vexim_virtual_delivery
rimas-kudelis: In fact, the transports should probably be split into multiple files. It seems our file is the only one which defines multiple transports. As for the naming, OK, you can leave it as is.

Z3po:
I'll open another pull request for that. Cleaner Commit history and stuff...

rimas-kudelis:
That wasn't necessary, but OK. :)

PR #33, opened 2015-03-06
Basic Bruteforce protection

Z3po:
I've implemented a basic bruteforce login protection so you can just login 11 times max within half an hour with a wrong password.
Probably there is a better solution to this and my php isn't the best so please carefully review and probably do it in a better way...

My Pull-Request is based on the still outstanding request for the debianic conf.d files.
Otherwise you could just cherry pick commit 11e4d91.

Let me know if there are any questions.


Z3po:
I'm not sure if we should merge these changes to the master branch as we're implementing new functionality here.
I guess a good Idea would be to "tag" the current release on the one hand and start working on a development branch with new features.
Probably start working on a 3.X ?

rimas-kudelis:
Hi @Z3po!
Firsty, you should have based your patch on current master. Sure, I can cherry-pick, but it would have been cleaner that way.
Secondly, regarding the screwup: I guess you meant to create `vexim`.`failed_logins` instead of `failed_logins`.`group_contents` :)
Thirdly, I think the number of failed logins should be configurable via a variable in variables.php. You can name it `$maxFailedLogins` or similarly (I just noticed that there is no consistency in how config variables are named, so you can choose whichever style you prefer).
Regarding your suggestion to tag it: maybe you're right. However, should we begin working on v3, I tend to think that first of all, it should be a full rewrite of the web app, as opposed to just some new functionality injected. Ideally, that rewrite would be properly versioned and self-upgrading the database as needed. Even if I give a version to the current state of Vexim before merging your PR, it won't be something very easily upgradable because we haven't been providing upgrade scripts or even documenting DB changes very well...

Z3po:
You're right with all of these.
I suggest we close that pull request for now and will add that as a feature to a new version vexim and start cleanup what we currently have.

rimas-kudelis:
okay

PR #34, opened 2015-03-09
adding debianic conf.d files

Z3po:
Now with a clean comit history. rebased to one commit to keep the history short.
all ideas from issue #32 are accepted and applied.
Hopefully I didn't lose anything :)


docs/debian-conf.d/main/00_vexim_listmacrosdefs
```diff
@@ -1,20 +1,36 @@
...
 MAIN_LOCAL_DOMAINS = MAIN_LOCAL_DOMAINS : ${lookup mysql{VEXIM_VIRTUAL_DOMAINS}} : ${lookup mysql{VEXIM_ALIAS_DOMAINS}}
 MAIN_RELAY_TO_DOMAINS = MAIN_RELAY_TO_DOMAINS : ${lookup mysql{VEXIM_RELAY_DOMAINS}}
 
-## If you don't want mail for localuser@virtualdomain delivered to the local
-## user, uncomment the domainlist line and adjust the appropriate routers in
-## /etc/exim4/conf.d/router/ (at least some of the ones with a prefix higher
-## than 250) to use this domainlist instead of local_domains.
-## NOTE: this only applies to the cases when the user does not exist in virtual
-## exim database. If the user exists, these routers 
-# domainlist local_user_domains = @ : localhost
```
rimas-kudelis: I think this should still be added somewhere (not necessarily here). I don't think local users should be getting e-mails sent to virtual domains.

docs/debian-conf.d/router/249_vexim_ditch_routers.erb
```diff
@@ -0,0 +1,128 @@
+
```
rimas-kudelis: What does .erb at the end of this file name mean?

docs/debian-conf.d/transport/30_vexim_ditch_spam_transport
```diff
@@ -0,0 +1,20 @@
+
+### transport/30_vexim_ditch_spam_transport
+#######################################
+
+ditch_spam_transport:
```
rimas-kudelis: This transport should get a "virtual_" prefix.

rimas-kudelis:
Sorry that I didn't mention some of these things previously. 249_vexim_ditch_routers.erb actually managed to escape my vision completely.
I can apply this patch anyway. My apologies, if this is getting annoying.

docs/debian-conf.d/transport/30_vexim_ditch_spam_transport
Z3po: I'm not sure..the normal "ditch_spam" blackhole router doesn't have that, too. Here's the Problem we have with "no naming rules". Still I'm adding the virtual here.

rimas-kudelis: You can add that prefix elsewhere as well. This way we will have naming rules and it will become very unlikely that our names would ever conflict with others.

docs/debian-conf.d/main/00_vexim_listmacrosdefs
Z3po: I wouldn't like to have that back in there...
* the comment is incomplete..it stops in the middle of a sentence
* there's no information what still needs to be done so it's useless anyhow
* I agree that information should go to the README file but it doesn't make sence to add it to the conf.d file directly. still routers need to be modified etc. If I come up with a better Idea to handle that I'll create a pull request :)

Z3po:
Naah no worries. I recognized that by myself earlier and forgot about it immediately. Sometimes my mind flips like pin-ball machine :)
Hopefully I didn't forget about anything now....

rimas-kudelis:
OK, I'm accepting all this, although I see a lot of room for improvement. In particular:
* we should really use a unique prefix for all our files, transports, routers etc.
* also, all files (the ones you added and those that existed before) should be reviewed with a critical eye and adjusted as necessary
* in addition, it would be cool to have automatic conversion between the full configure file and debian-style conf.d. This way these two configuration storages could be easily kept in sync. Perhaps it would actually make sense to adopt Debian-style layout along with default Debian config files as the base format, and use the update-exim4.conf script to generate plain configure file...

rimas-kudelis:
One of the things that should probably be revisited is indeed the `local_part_suffix` settings of our routers, for example. I know I asked you to uncomment them, but now I'm thinking that perhaps a better thing would have been to simply comment them out in configure file as well.

PR #35, opened 2015-05-26
'clear' field

kvsmirnov:
'clear' field does not changed or updated in sql requests to database

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -57,8 +57,8 @@
   $aliasto = preg_replace("/[', ']+/", ",", $_POST['smtp']);
   if (alias_validate_password($_POST['clear'], $_POST['vclear'])) {
     $query = "INSERT INTO users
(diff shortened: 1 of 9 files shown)


Udera:
We are planning to remove the clear field completely. It makes no sense to store an encrypted password along with the clear-text password. If you want to use clear-text passwords, you can choose 'clear'-text as crypto-method:
https://github.com/avleen/vexim2/pull/23

rimas-kudelis:
Exactly. This patch basically undoes the commits from #23.

@kvsmirnov if you are not using the `clear` field in your setup, but are simply worried that it is not being synchronized with the `crypt` field, you should just delete that field from the table (it should have been done automatically, but sadly, we don't have proper table versioning and migration scripts).

On the other hand, if your setup requires you to use cleartext passwords, then having an encrypted copy of them adds very little security. In that case, you should probably just use that unencrypted password in all parts of your setup, although I would highly recommend to instead substitute the software which does not support encryption with one that does.

PR #36, opened 2015-06-08
Fixed some bugs, added Russian translation

kvsmirnov:
Clear text passwords need to have no limit in the methods of authorization, but work with it was broken.
Two lines in project was'nt have translation.
Added Russian translation.
Fixed displaing national symbols. Default encoding in gettext is now UTF-8


rimas-kudelis:
As mentioned previously in #35, we do not want to reintroduce the `clear` field.
If you want to submit just the Russian localization, this is more than welcome. Bug fixes and code style fixes are more than welcome. But the `clear` field is definitely not.
Also, bug fixes that are isolated from each other have better chances to be accepted into the repository. For example, I would accept the missing quotes fix to `adminuserblocksubmit.php` without thinking much, but I'm not so sure about setting locale to *.UTF-8 explicitly.

Issue #37, opened 2015-06-08
Remove remaining notions of the `clear` password field


rimas-kudelis:
Apparently, in #23 not all notions of the `clear` field were removed. This may have lead us to #35 and #36, where patches have been proposed which were supposed "fix" non-handling of that field in the code (basically by undoing #14).
We should get rid of these remnants.
@Udera, would you perhaps like to do this?

rimas-kudelis:
The files that seem to be still affected are under `docs` and `setup`:
* `docs/configure`
* `docs/database-spec.txt`
* `docs/clients/mysql-popper.conf`
* `setup/create_db.pl`
* `setup/mysql.sql`
* `setup/pgsql.sql`

Udera:
I can have a look into it, removing it shouldn't be a problem.
I was wondering what happens with users who want to update their setup? This requires more migration steps (some might want to change the hashing algorithm as md5 is probably not state of the art).

rimas-kudelis:
AFAIK, Exim4U has written some migration scripts, we should see if we could reuse that.

But frankly speaking, I'm not sure the current Vexim codebase is worth improving. Coding standards and practices have improved a lot since it was written, so it might be (and I think it is) better to just rewrite it from scratch, using a contemporary approach (modern PHP5 framework, separation of php code from templates etc.). Ideally, the framework would have built-in support for migrations (and migration generations), so we could concentrate on writing and improving business logic instead of dealing with a constantly increasing amount of database versions we have to migrate from.

In #33, @Z3po suggested something similar: his idea is that we should ship this release and start working on a new one. I think, this could be our chance to start from the ground up. Would you be interested in that?

Udera:
You are probably right with a new start and your ideas sound good. I would be happy to help out but I have to say that I'm not a software developer.

Z3po:
I'd vote for that, too.
Let's build up a "clean" stable release (probably with the clear-field back in.
I'm no web developer either but would be happy to add some lines of code when having time for it...

rimas-kudelis:
I don't think we really want the `clear` field back. Unless I misunderstood what you meant. :)

Z3po:
So we have 2 issues here:
* nobody likes to have the clear field. I agree with that, but:
* we've introduced the clear field and don't have any migration scripts ready for removing it within the already existing mysql tables...

So I see the following approaches:
* freeze with the clear removed and have "broken" database setups
[...]

PR #38, opened 2015-06-10
Some enhancements and bug fixes. Added Russian translation

kvsmirnov:
1. 'clear' & 'crypt' field replaced by one field 'password'. To upgrade
from previous version you need to see upgarding.txt in docs
2. Spaces and tabs at line end deleted
3. Added Russian translation. Report ro skv@ivsor.net if you found
mistakes
4. Boxes Antivirus and Antispam checked by default


rimas-kudelis:
Hi,

Thanks for your contribution, but can you please make separate commits and separate pull requests for such different changes? That would greatly increase the likeliiness of some of them getting in the codebase really quickly.

Also, I don't think we should replace both password fields with a new one.This would have been a good choice, if we didn't have to care much about legacy or if there was any substantial benefit from such change. But we do have legacy, and renaming the fields won't help anyone, so I think we should just leave the default field as it is (`crypt`). I did want to get rid of the all mentions of the `clear` field in #37, but after discussing with @Udera, I'm not so sure even about that anymore. I suppose we will have to ask our user to do some manual migration anyway, but I'm not convinced that now is the right time for that.

By the way, if you have anything to add to the discussion on that bug, feel free to do so.

PR #39, opened 2015-06-16
Salt generation: Replace uniqid() with a more secure function from the openssl-library.

Udera:
Some inspiration from https://github.com/Exim4U/src/blob/master/home/exim4u/public_html/exim4u/config/functions.php


rimas-kudelis:
I'm not so sure that it's necessary (IMO, salt doesn't really need to be random, it just has to vary), but why not.

PR #40, opened 2015-06-17
Add SSL/TLS-certificate in the config-file because we require TLS-aut…

Udera:
…hentication.
We ask for authentication of users via TLS. Therefore it is necessary to activate TLS in the configuration-file. I still have a single-file-configuration running and we should provide a working example.
The debian config-file (conf.d/main/03_exim4-config_tlsoptions) already has TLS enabled by default (with auto-generated certificates?).

### DIFF
--- a/README
+++ b/README
@@ -257,6 +257,18 @@ Exim configuration:
     in /var/tmp, /usr/lib, or any number of other places. If yours isn't
     /tmp/mysql.sock, you will need to set this.
   ----------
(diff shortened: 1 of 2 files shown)


rimas-kudelis:
Thanks! I had forgotten it, but TLS is probably the main reason why I considered our configure file as not working. If you have any further patches to the configure file, feel free to share them!

PR #41, opened 2015-06-18
Router update

Udera:
In the debian-configs, there is a more flexible spam-handling that allows you to either drop spam mails or deliver them to a specific sub-folder. I added the router and transports from there.
Now it's more or less the same (I only checked the existing transport, I didn't run a real diff).


Udera:
sorry, it's a mess. I created a new branch https://github.com/avleen/vexim2/pull/42

PR #42, opened 2015-06-18
update single exim single config file

Udera:
Now, it should be a cleaner request. Put same spam-routes and -transports into single config-file.


rimas-kudelis:
Thanks! I haven't looked inspected these patches thoroughly, but I trust that you just copied and pasted these blocks from Debian files, so there's no reason for us not to have them in the main file as well.

PR #43, opened 2015-06-18
add fields that are already in the exim-routers

Udera:
the spam_drop-field is used here: https://github.com/avleen/vexim2/blob/master/docs/debian-conf.d/router/249_vexim_ditch_routers#L49
even it didn't exist before!


setup/mysql.sql
```diff
@@ -59,6 +59,7 @@ CREATE TABLE IF NOT EXISTS `vexim`.`users`
 	on_piped         bool                    NOT NULL  default '0',
 	on_spamassassin  bool                    NOT NULL  default '0',
 	on_vacation      bool                    NOT NULL  default '0',
+	spam_drop		bool                    NOT NULL  default '0',
```
rimas-kudelis: can you please fix the alignment here?

setup/create_db.pl
```diff
@@ -221,6 +221,7 @@ sub create_mysqltables {
 	on_piped bool NOT NULL default '0',
 	on_spamassassin bool NOT NULL default '0',
 	on_vacation bool NOT NULL default '0',
+	spam_drop bool NOT NULL default '0',
```
rimas-kudelis: Can you add this field to the commented out block in `create_postgrestables` (see Line 330)?

Udera:
@rimas-kudelis thanks, I missed the commented part.

rimas-kudelis:
It seems we also don't have the UI which would set this flag. Could you please add that as well?

Udera:
I have opened an issue here (https://github.com/avleen/vexim2/issues/44) that we don't forget to fix this. Either I put this in a new pull request or if you don't merge it yet, I will attach it to this one.

setup/mysql.sql
```diff
@@ -59,6 +59,7 @@ CREATE TABLE IF NOT EXISTS `vexim`.`users`
 	on_piped         bool                    NOT NULL  default '0',
 	on_spamassassin  bool                    NOT NULL  default '0',
 	on_vacation      bool                    NOT NULL  default '0',
+	spam_drop	 bool                    NOT NULL  default '0',
```
rimas-kudelis: This still should use spaces, except at the beginning of the line.

Udera:
I added the spam_drop function to the web-interface.

vexim/userchange.php
```diff
@@ -47,60 +47,119 @@
...
+          ?>
+  	  <tr>
+  	    <td><?php echo _('Maximum message size'); ?>:</td>
+  	    <td>
+            <input type="text" size="5" name="maxmsgsize"
+              value="<?php echo $row['maxmsgsize']; ?>" class="textfield">Kb
+	    </td>
+  	  </tr>
+  	  <tr>
+  	    <td><?php echo _('Vacation on'); ?>:</td>
```
rimas-kudelis: I think "Vacation enabled" was a better message.

vexim/userchange.php
```diff
@@ -47,60 +47,119 @@
...
+	    </td>
+  	  </tr>
+  	  <tr>
+        <?php if (function_exists('imap_qprint')) { ?>
+  	    <td><?php echo _('Vacation message'); ?>:</td>
+  	    <td>
+            <textarea name="vacation" cols="40" rows="5" class="textfield"><?php print imap_qprint($row['vacation']); ?></textarea>
+          </td>
+        <?php } else { ?>
+  	    <td><?php echo _('Vacation message (ASCII only!)'); ?>:</td>
```
rimas-kudelis: This could probably say "ASCII only or QP-encoded!"

rimas-kudelis:
Thanks @Udera!
Next time please avoid big cosmetic changes in unrelated pull requests. They make it more difficult to read a diff.
I left two comments on the strings, but I'm merging your PR in.

Udera:
@rimas-kudelis I first made the changes to adminuseradd.php and the style there is different. I didn't want to rewrite everything. And the style with putting all html-code into print-commands is not very readable either.

rimas-kudelis:
Well, I can't disagree on that. This file was really irregular.

rimas-kudelis:
Oops, a bug slipped in this patch. I've just pushed a fix: df0bfca .

PR #45, opened 2015-08-11
Minor modification of example config file to configure against OpenSSL cy…

ikonia:
…pher rather than GNUTLS, distro link exim against OpenSSL now rather than GNUTLS so this makes more sense

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -275,7 +275,7 @@ tls_privatekey = /etc/exim4/exim.key
 tls_dhparam = /etc/exim4/dhparam.pem
 tls_require_ciphers = ${if =={$received_port}{25}\
 {NORMAL:%COMPAT:!VERS-SSL3.0}\
-{SECURE128}}
+{MAX}}
 
 
 ######################################################################


rimas-kudelis:
Does this work with both libraries?

ikonia:
No, ssl only,

The one that is there currently only works with gnutls. As the distros all seem to links against ssl it made sense to me to make this the default option. He longer term I would expect to have multiple lines in for distros/options and you just uncomment what you want to use.

Matt

rimas-kudelis:
The library is called OpenSSL, not just SSL. :)

The [Exim manual](http://www.exim.org/exim-html-current/doc/html/spec_html/ch-encrypted_smtp_connections_using_tlsssl.html#SECTreqciphssl) offers the following config for OpenSSL:
```
# OpenSSL variant; see man ciphers(1)
tls_require_ciphers = ${if =={$received_port}{25}\
                           {DEFAULT}\
                           {HIGH:!MD5:!SHA1}}
```
and the following for GnuTLS:
```
# GnuTLS variant
tls_require_ciphers = ${if =={$received_port}{25}\
                           {NORMAL:%COMPAT}\
                           {SECURE128}}
```

Perhaps we should use these settings? Also, perhaps we should provide both blocks, but have one of them commented out (or maybe even both)?

SvenBunge:
We've started with the GnuTLS variant from the documentation. Now we're using:
```
tls_require_ciphers = ${if =={$received_port}{25}\
    {NORMAL:%COMPAT:-ARCFOUR-128:-ARCFOUR-40:-MD5:-VERS-SSL3.0}\
    {SECURE128:+AES-256-CBC:+CAMELLIA-256-CBC:-ARCFOUR-128:-ARCFOUR-40:-MD5:-VERS-SSL3.0}}
```
So we use normal and compatible (also no-)ciphers on the default SMTP port but deny RC4 and SSL3. Google is using RC4 as default cipher for example. If you deny it they fallback to AES. Really strange behavior :-)
For the SMTP delivery port we only allow secure ciphers and we've added some 256bit ciphers and deny RC4 and SSL3 as well.

We're currently using Debian Wheezy. Maybe this helps. 
(But the settings is not compatible with OpenSSL.)

rimas-kudelis:
@ikonia in fact, Debian seems to link against GnuTLS.
Like I said, I think it would be nice to have two examples, one or both commented out, but I'm waiting for further input and/or code here.

Udera:
I added the last modification with disallowing SSLv3. I didn't try with RC4. The problem in general is that we still allow a fall-back to clear-text.

As Debian links against GnuTLS (assuming most people here use Debian), I would keep this as default, but add the OpenSSL version for other systems.

I'm not sure what way we should prefer: either provide a nearly complete set of configuration files that vexim can be run quickly and without cumbersome user interaction (good default configuration), or only the vexim-part and keep all other stuff to the distribution (encryption-settings, ports, certificates, ...).

rimas-kudelis:
2015-09-10 02:11, Udera rašė:
>
> I added the last modification with disallowing SSLv3. I didn't try
> with RC4. The problem in general is that we still allow a fall-back to
> clear-text.
>

Where?

> As Debian links against GnuTLS (assuming most people here use Debian),
> I would keep this as default, but add the OpenSSL version for other
> systems.
>

Agreed. Like I said before, I think both examples should be there
(ideally, both resulting in equivalent functionality), and one or both
of them should be commented out.

> I'm not sure what way we should prefer: either provide a nearly
> complete set of configuration files that vexim can be run quickly and
> without cumbersome user interaction (good default configuration), or
> only the vexim-part and keep all other stuff to the distribution
> (encryption-settings, ports, certificates, ...).
>

The question is not only what we want to ship, but also what we can
ship. If we can create a nicely working generic Exim config file, then
why not, but if we can't or don't want to maintain it, then shipping
just the relevant parts might be more feasible.

Udera:
The current pull request doesn't work for me, it does know the "MAX" value.
@SvenBunge proposal seems to work.
https://bettercrypto.org/static/applied-crypto-hardening.pdf says:
>tls_require_ciphers needs a GnuTLS priority string instead of a cipher list. It is recommended
to use the defaults by not defining this option. It highly depends on the version of GnuTLS used. Therefore it is not advisable to change the defaults.
On this site you can test the ciphers: https://ssl-tools.net/mailservers

rimas-kudelis:
Hm, maybe we should define a macro at the top of our config file, which we could then use to differentiate OpenSSL vs. GnuTLS?

rimas-kudelis:
By the way, bettercrypto.org has a good point regarding this:
> accept all cipher suites, as the alternative would be to fallback to clear text transmission

Perhaps we should just comment this setting out in our config?

Udera:
At least SSLv2 should be disabled: https://www.openssl.org/blog/blog/2016/03/01/an-openssl-users-guide-to-drown/

Most major mails servers should support a version of TLS 1.x. You can check the server settings with:
https://ssl-tools.net/mailservers

I would prefer a setting that can be used on most of the systems and which is not too bad (RC4/SSLv2/3 disabled). We can provide better settings of some specific versions we tested (with PFS and such stuff) in our wiki.

hotstepper13:
@SvenBunge
Thank you very much. Yesterday I spend a whole day in searching for a working configuration and only saw the the most tutorials assuming the openSSL variant. Ubuntu (as it is more like debian) also uses the GnuTLS variant.

@all
The problem in mentioning the correct cipher suite is not only if the connection is secured or not. You need to be compatible to all those other providers out there. I agrre with @Udera because it makes no sense to configure something secure and opening allowing others to make use of vulnerabilities we only have for that reason.

Example: Use plain text and someone can intercept the communication, getting mail content and a working user account. Using some faulty ssl implementation and be target of, lets say Heartbleed and someone could get much more stuff like cert keys and data from other users, without needing to act as man in the middle. Therefore to use ssl bugs is easier that to intercept plain communication without access to the server. (Just my two cents!)

BTW: Just to highjack this thread: If anyone has a working example of a PFS configuration for the GnuTLS variant it would be great to have it shared :)

rimas-kudelis:
@hotstepper13:
> Therefore to use ssl bugs is easier that to intercept plain communication without access to the server. (Just my two cents!)

I'm afraid this logic is a bit flawed. Heartbleed was an implementation bug, and it was fixed as soon as it was discovered. Meanwhile, at least to my understanding, all "insecure" cyphers are simply "not secure enough" anymore. They might be weak and crackable by today's standards, but I find it really hard to believe that spying on (weakly) encrypted traffic can be any easier than spying on plaintext traffic.

@Udera:
> At least SSLv2 should be disabled: https://www.openssl.org/blog/blog/2016/03/01/an-openssl-users-guide-to-drown/

Yes of course, it should be disabled. But I believe we aren't the ones who should be disabling it. Instead, it should be done by developers or vendors of the libraries providing us with encryption. I mean, these people are encryption gurus, not us. They should know better and I'm pretty sure they do. I doubt it's feasible to keep adjusting the configuration files we distribute whenever a new vulnerability surfaces, because these vulnerabilities get fixed upstream really quickly, so by the time our user will be installing Vexim with workarounds for bug X, these workarounds will most likely be already irrelevant, because either the underlying problem will have been fixed, or the affected cipher/whatever disabled in the library itself.

That's what I think. Correct me if I'm wrong.

Udera:
> Yes of course, it should be disabled. But I believe we aren't the ones who should be disabling it. Instead, it should be done by developers or vendors of the libraries providing us with encryption. I mean, these people are encryption gurus, not us. They should know better and I'm pretty sure they do. I doubt it's feasible to keep adjusting the configuration files we distribute whenever a new vulnerability surfaces, because these vulnerabilities get fixed upstream really quickly, so by the time our user will be installing Vexim with workarounds for bug X, these workarounds will most likely be already irrelevant, because either the underlying problem will have been fixed, or the affected cipher/whatever disabled in the library itself.
> 
> That's what I think. Correct me if I'm wrong.

True, should be. That's the reason I don't want to enforce perfect forward secrecy (which might not be available on older systems). But if you have an older system, do maintainers deprecate old ciphers? Debian wheezy, or ubuntu 14.04 perhaps still run on SSLv2/3 per default.

hotstepper13:
@rimas-kudelis 
Thant you for your thoughts. The usage of hartbleed on similar attacks only need remote access to the server which means the ability to sending requests to the port that use ssl. Under some special circumstances the server was sending other data in its response allowing others to see data from different sessions, server configurations and so on.

On the other hand, to sniff any type of traffic you need access to the target server or you need access to core network infrastructure (like routers, loadbalancers) or you need the victim to use a different name resolution in order to reroute traffic to some system that is under your control.

That is the reason why I am saying that the usage of SSL Bugs like Heartbleed are easier than sniffing / capturing plain traffic from a user / connection.

Regarding the "is it our job?" part, i think the same goes to spamassassin and clamav. This repo / project is orginated to a virtual user management for exim. But you didnt only cover vexim, instead you give the people good and useful advices how to setup their mail infrastructure to make it work safer and better.

Therefore I think you guys should setup the cipers in a good way and adding some commented "activate this instead of the other if you encounter trouble" configuration.

rimas-kudelis:
@Udera: I would like to assume that every OS vendor which is serious about security would quickly push security updates and disable things that are found out to be inherently insecure. E.g. if SSLv3 is found to be insecure by design, then I really hope Debian would push security updates and disable SSLv3 on all their supported OS versions.

@hotstepper13: same applies for Heartbleed. I don't see how fixing potential Heartbleed issues on OUR side can be relevant today. If someone is using Vexim on an obsolete OS without security updates (or simply using it on a supported OS without bothering to update it for years), then this is insecure by definition, and it's that person's job to actually update their machine. Furthermore, if they're using Vexim on an unsupported/not updated OS, it's quite likely that they aren't updating Vexim either, in which case, releasing a fix targeting them is pointless. On the other hand, if someone uses Vexim on a supported OS, then I trust that the maintainers of that OS will take care to push fixes for things like Heartbleed to their users in a timely fashion, again, making it pointless for us to try to work it around in our product only.

Regarding Spamassassin and Clamav, the things are a bit different, because 1) we integrate with these products and provide some configuration options related to them, and 2) we integrate only with these particular products at the moment (e.g., we don't have alternative configuration examples for Kaspersky or whatever other AV engine you may have in mind). That makes things a bit easier. But even then, I don't think we touch these products that much, do we?

Udera:
@rimas-kudelis : I justed checked on a fully patched ubuntu 14.04 (it's not the newest released but still supported for some time). Without specifying the ciphers at all, it supports SSLv3 and RC4!

rimas-kudelis:
Honestly, assuming these are really bad, that sounds like a bug in Ubuntu to me. I wonder if it has been reported to them and if yes, whether Ubuntu is going to fix it, or if they have reasons not to.

Udera:
I don't think that they fix such things. SSLv2 or RC4 is not insecure 
because of a wrong implementation or an error code, it's by design. And 
they don't change default settings of released version because this 
could break existing setups. If you have an old enough Linux, you'll 
find plenty of settings that are today considered insecure (sha1 on ssl 
certificates, SSLv2/3, ...).

rimas-kudelis:
@Udera, hm, you might be right.

Udera:
I played a bit with debian 8 and RC4 is used by default. So the version in https://github.com/vexim/vexim2/pull/45#issuecomment-130092012 seems to be a better choice. In newer gnutls-version (e.g. in debian 8) you can even enforce perfect forward secrecy but anyone can improve it. So I would take this proposal, it works at least on debianic systems.

Can anybody test with FreeBSD?

Udera:
Finally, I used this proposal https://github.com/vexim/vexim2/pull/45#issuecomment-130092012 to write a follow up of this topic.

PR #47, opened 2015-08-17
Code clean-up

Kami-no:
Clean code with 2 spaces ident and no trailing spaces

### DIFF
--- a/vexim/adminaliaschange.php
+++ b/vexim/adminaliaschange.php
@@ -84,8 +84,7 @@ if ($sth->rowCount()) {
               able to log in, or if the Alias is the admin account'); ?>)
                     </td>
                 </tr>
-                <tr
-                    >
+                <tr>
                     <td><?php echo _('Verify Password'); ?>:</td>
                     <td>
                         <input name="vpassword" type="password" size="30" class="textfield">
--- a/vexim/adminaliaschange.php
+++ b/vexim/adminaliaschange.php
@@ -5,7 +5,7 @@ include_once dirname(__FILE__) . '/config/functions.php';
 include_once dirname(__FILE__) . '/config/httpheaders.php';
 $query = "SELECT localpart,realname,smtp,on_avscan,on_spamassassin,
     admin,enabled FROM users
-	WHERE user_id=:user_id AND domain_id=:domain_id AND type='alias'";
+    WHERE user_id=:user_id AND domain_id=:domain_id AND type='alias'";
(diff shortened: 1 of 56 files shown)


rimas-kudelis:
I would prefer PSR-compliant code (it uses 4 spaces per indentation step, if I remember correctly), but do you think it's really worth it? This codebase is anything but clean either way...

Kami-no:
PSR - no problem. I prefer it too but in your code 2 spaces are used more often. I'll change it. I'm going to clean all code anyway. Clean code helps to read the code.

I'm interested in vexim at CentOS for the project in my company. I'd like to make some blacklists and whitelists.

I just offer my changes to upstream.

Kami-no:
Nice clean code now. By the way, I've found several times code after die. IDE noticed that it will never be executed. Next set of updates getting ready but they're not for this pull.

vexim/adminaliaschange.php
```diff
@@ -1,144 +1,145 @@
...
-		<?php 		
-			# end of the block editing an alias within the domain
-		}  
-		?>	
-    </div>
-  </body>
+                    </td>
+                </tr>
+                <tr
+                    >
```
rimas-kudelis: You seem to have missed this.

vexim/adminfailchange.php
```diff
@@ -1,66 +1,67 @@
...
+<body onLoad="document.failchange.localpart.focus()">
+<?php include dirname(__FILE__) . '/config/header.php'; ?>
+<div id="menu">
+    <a href="adminfail.php"><?php echo _('Manage Fails'); ?></a><br>
+    <a href="adminfailadd.php"><?php echo _('Add Fail'); ?></a><br>
+    <a href="admin.php"><?php echo _('Main Menu'); ?></a><br>
+    <br><a href="logout.php"><?php echo _('Logout'); ?></a><br>
+</div>
+<div
+="Forms">
```
rimas-kudelis: Obviously an accident.

rimas-kudelis:
But seriously, I really doubt that fixing indentation makes this code any better. I think I see it as being in maintenance phase where only bugfixes are to be applied when necessary. If you want to remove closing PHP tags or fix bugs in the code, that's OK, but I don't think that changing indentation of 95% of the code makes sense at its current state.
Oops, I'm repeating myself... :(

Kami-no:
No problem to me if you would not accept the Pull Request.

Issue #50, opened 2015-08-24
Mail storage dir /var/spool/vexim ?


Kami-no:
Readme.md
* row 97 /var/spool/vexim/
* row 334 /usr/local/mail/

variables.php
* row 57 $mailroot = "/usr/local/mail/";

Confusing...

https://github.com/avleen/vexim2/pull/51

Udera:
Oh, that is strange indeed. In debian, I have the mail of local users delivered to /var/mail/$localpart. /var/spool is only for temporary usage (e.g. exim queue).

We fixed some stuff over the last months, we need to review the documentation and make sure to have a working version.

rimas-kudelis:
This is fixed by 7aa018aff625ca3400da30baf8f387dc70ebc42f.

PR #51, opened 2015-08-31
Default mail storage same as in manual

Kami-no:
No need to remember that if you use the manual.


Udera:
````Spool for tasks waiting to be processed, e.g., print queues and outgoing mail queue.````
https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

I don't think you want to store your emails at this location.

Kami-no:
Well, /usr also is not an option...
Secondary hierarchy for '''read-only user data'''; contains the majority of (multi-)user utilities and applications.[6]

/var/mail or /var/mbox

Most Linux distributives (including Debian and RHEL) prefer /var/www for web-servers' home directory. /var is the optimal location for such data.

Udera:
/var/mail is already used in Debian as mailbox for system users.
We could use /var/vmail, /var/mbox, or /var/vexim ... in the end the manual only suggests a location

rimas-kudelis:
Just FYI, I myself use /var/mail/vexim for this. Debian using /var/mail hasn't got in my way yet, but if we want to be cautious, /var/vmail sounds just fine to me.

Kami-no:
As you wish...

Kami-no:
BTW you've marked "duplicate" the wrong thread.

rimas-kudelis:
Please squash the commits, if you don't mind.
Regarding marking the thread duplicate: it's just a flag anyway, but I thought the other issue describes this issue better. Once the PR is merged, both issues will be closed anyway (especially if you say `fixes #50` in your commit message :) ).

rimas-kudelis:
Still two commits?

rimas-kudelis:
Your changes have been merged. Thank you! :)

PR #52, opened 2015-08-31
Unreachable statements

Kami-no:
There is a «die» instruction in the rows above.


rimas-kudelis:
Actually, I think this should be fixed by making the statements reachable instead, don't you?

Kami-no:
If you check the code, you'll see the IF with the same code as I offer to remove in ELSE section.
All cases are just duplicates...

Udera:
There is something wrong. But we can't just drop the statements. If you disable spamassassin for a given domain, you need to set on_spamassassin=0 for all users of this domain. Currently this is not done. @Kami-no can you have a look at it?

Kami-no:
Now I see... Ok, it will be done.

rimas-kudelis:
@Kami-no: could you squash these commits, please?

rimas-kudelis:
Also, userchangesubmit.php has these premature die() statements as well. Would you please fix them the same way you fixed sitechangesubmit.php?

rimas-kudelis:
This is still two commits, and create_db.pl is now broken. Would you like me to help you tame git? :)

Kami-no:
Yes please, cycled many times through squash - incompatible - merge from original - squash - incompatible
No progress here and in storage...

rimas-kudelis:
When you are rebasing, you should resolve arising commit conflicts manually, you can't just commit files with conflicts as if they are OK.
And when you resolve conflicts, you should continue the rebase by running `git rebase --continue`. I guess all this mess happened because you didn't do one or both of these things.
I've committed a proper change on your behalf and am closing this PR, but you may still want to fix `userchangesubmit.php`, as that contains lots of `die()` statements which prevent further code from running. Please do that in a clean tree and issue a new PR. Thanks!

PR #53, opened 2015-09-09
Cleanup clear-field

Udera:
Remove "clear" field from database and use crypt-field only. First, I wanted to include a more modern password scheme (SHA512) into this pull request. But there need to be more modifications (perl-script, vexim-phps, exim/dovecot-config, documentation), so I decided to put this in a separate pull request.


rimas-kudelis:
Thank you!

PR #55, opened 2015-09-13
Markdown readme

Kami-no:
Better visual Readme (and project homepage)


rimas-kudelis:
Please rebase on latest master. Also, I suppose the original README is no longer necessary if we move its content over to README.md, is it?
Either way, when updating the PR, please retain the number of commits in the PR at one, as it is now.

rimas-kudelis:
Ouch, I suppose you squashed wrong commits, that's why your diff is so uselessly big now. Would you like to fix this PR on your own, or shall I just commit the change on your behalf?

Kami-no:
I'll fix it.

PR #60, opened 2015-09-17
Small fix

Udera:
just two small submits to put the new user name and to have the same ditch_-routers.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -480,10 +480,11 @@ ditch_maxmsgsize:
 		and domain = '${quote_mysql:$domain}' \
 		and users.maxmsgsize > 0 \
 		and users.domain_id=domains.domain_id}{${value}K}fail} Kb.
-  local_part_suffix = -*
-  local_part_suffix_optional
+#  local_part_suffix = -*
+#  local_part_suffix_optional
   retry_use_local_part
 
+# malware drop router
 ditch_malware:
   driver = redirect
   allow_fail
@@ -543,8 +544,8 @@ ditch_hdrmailer:
   			and domains.domain = '${quote_mysql:$domain}' \
 			and domains.domain_id=blocklists.domain_id \
 			and users.user_id=blocklists.user_id}}}{1} {yes}{no}}
-  local_part_suffix = -*
-  local_part_suffix_optional
+#  local_part_suffix = -*
+#  local_part_suffix_optional
   retry_use_local_part
 
 ditch_hdrto:
@@ -558,8 +559,8 @@ ditch_hdrto:
   			and domains.domain = '${quote_mysql:$domain}' \
 			and domains.domain_id=blocklists.domain_id \
 			and users.user_id=blocklists.user_id}}}{1} {yes}{no}}
-  local_part_suffix = -*
-  local_part_suffix_optional
+#  local_part_suffix = -*
+#  local_part_suffix_optional
   retry_use_local_part
 
 ditch_hdrfrom:
@@ -574,8 +575,8 @@ ditch_hdrfrom:
 			and domains.domain_id=blocklists.domain_id \
 			and users.user_id=blocklists.user_id}}}{1} {yes}{no}}
   local_part_suffix = -*
-  local_part_suffix_optional
-  retry_use_local_part
+#  local_part_suffix_optional
+#  retry_use_local_part
 
 ditch_hdrsubject:
   driver = redirect
@@ -588,8 +589,8 @@ ditch_hdrsubject:
   			and domains.domain = '${quote_mysql:$domain}' \
 			and domains.domain_id=blocklists.domain_id \
 			and users.user_id=blocklists.user_id}}}{1} {yes}{no}}
-  local_part_suffix = -*
-  local_part_suffix_optional
+#  local_part_suffix = -*
+#  local_part_suffix_optional
   retry_use_local_part
 
 virtual_vacation:
--- a/docs/configure
+++ b/docs/configure
@@ -77,7 +77,7 @@ ALIAS_DOMAINS = SELECT DISTINCT alias FROM domainalias WHERE alias = '${quote_my
 domainlist local_domains = @ : example.org : ${lookup mysql{VIRTUAL_DOMAINS}} : ${lookup mysql{ALIAS_DOMAINS}}
 domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
 hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
-trusted_users = vexim:www-data
+#trusted_users = www-data
 
 #Debian: hide mysql_servers = localhost::(/var/run/mysqld/mysqld.sock)/vexim/vexim/CHANGE
 #hide mysql_servers = localhost::(/tmp/mysql.sock)/vexim/vexim/CHANGE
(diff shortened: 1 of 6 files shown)


docs/configure
```diff
@@ -478,10 +480,11 @@ ditch_maxmsgsize:
 		and domain = '${quote_mysql:$domain}' \
 		and users.maxmsgsize > 0 \
 		and users.domain_id=domains.domain_id}{${value}K}fail} Kb.
-  local_part_suffix = -*
-  local_part_suffix_optional
+#  local_part_suffix = -*
+#  local_part_suffix_optional
```
rimas-kudelis: I'm not sure we should comment these out. Perhaps we should add them to the Debian files instead?

Udera: People from old setups might be using this option, so this is perhaps the better option.

docs/configure
```diff
@@ -77,8 +77,9 @@ ALIAS_DOMAINS = SELECT DISTINCT alias FROM domainalias WHERE alias = '${quote_my
 domainlist local_domains = @ : example.org : ${lookup mysql{VIRTUAL_DOMAINS}} : ${lookup mysql{ALIAS_DOMAINS}}
 domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
 hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
-trusted_users = avleen:www
+trusted_users = vexim:www-data
```
rimas-kudelis: I wonder if we need this either. Exim docs say, that " If this option is set, any process that is running as one of the listed users is trusted", but I don't think anything is supposed to be run under the vexim user. Furthermore, [here's what trusted users can do](http://www.exim.org/exim-html-current/doc/html/spec_html/ch-the_exim_command_line.html#SECTtrustedadmin), and again, I don't think vexim needs this, do you?

On my working Debian system the following are the defaults that I never bothered to change. And everything works with them:
```
sonata:/etc/exim4/conf.d# grep -iR trusted_users
main/02_exim4-config_options:.ifndef MAIN_TRUSTED_USERS
main/02_exim4-config_options:MAIN_TRUSTED_USERS = uucp
main/02_exim4-config_options:trusted_users = MAIN_TRUSTED_USERS
sonata:/etc/exim4/conf.d#
```
I think we can safely remove this option, or at least comment it out, leaving only www-data in the commented out list. I'd actually be in favor of removal, because the smaller the config we provide, the easier it is to extend and improve for the end-user.

docs/configure
rimas-kudelis: Please explain. :)

I just had an idea: would it be possible to do like Debian does and define a macro/variable/whatever at the top of `configure` which could be easily changed, and which would switch these options on or off? Something like this:

```
VEXIM_LOCALPART_SUFFIX = -*

...

.ifdef VEXIM_LOCALPART_SUFFIX
local_part_suffix = VEXIM_LOCALPART_SUFFIX
local_part_suffix_optional
.endif
```

rimas-kudelis: By the way, I myself noticed that I use `+*` for localpart suffix. At least my impression is that this is used more often than `-*`. Perhaps we should change our default?

docs/configure
Udera: That there might be problems with cronjobs, quota-messages, ... if the mail addresses have not been configured properly. If we change this, we should keep a document with these changes that allows you to check possible problems.

docs/configure
Udera: Ok, let's keep the local_part_suffix in the configuration, I will go through the conf.d/-files and add these.

Is + the more common suffix? Or we use both?
````
local_part_suffix = +* : -*
````

File: docs/configure (hunk 2 above)
Comment (rimas-kudelis) [inline docs/configure]: Well again, no process runs under vexim user, so by extension, no process running under that user would send any e-mails. I agree we can document this somewhere. Perhaps in the release notes?

File: docs/configure (hunk 1 above)
Comment (rimas-kudelis) [inline docs/configure]: I thought I replied to this, but I can't find that reply.
I think one suffix should be enough by default. We could even leave it commented out, so that the administrator would have to knowingly enable this feature.
I think the plus sign might be more common in this role, but I don't have any data to back this opinion.

Comment (syscrusher1709) [inline docs/configure]: On Sep 17, 2015 12:37 PM, Udera <notifications@github.com> wrote:
>
> In docs/configure:
>
> > @@ -478,10 +480,11 @@ ditch_maxmsgsize:
>
> >  		and domain = '${quote_mysql:$domain}' \
>
> >  		and users.maxmsgsize > 0 \
>
> >  		and users.domain_id=domains.domain_id}{${value}K}fail} Kb.
>
> > -  local_part_suffix = -*
>
> > -  local_part_suffix_optional
>
> > +#  local_part_suffix = -*
>
> > +#  local_part_suffix_optional
>
> Ok, let's keep the local_part_suffix in the configuration, I will go through the conf.d/-files and add these.
>
> Is + the more common suffix? Or we use both?
>
> local_part_suffix = +* : -*
I have often seen hyphens within an address local part, not as a suffix. In my experience the only common sufffix delimiter is the plus sign.
By the way, I'm new to this list but have used Vexim for ages. I signed on as I'm preparing to rebuild/update my servers. I was afraid Vexim would have gone dormant, but I'm really happy to see such a vibrant community! I did some local patches to add features. I'll review these against the current version and contribute any that are still useful.
Scott

**Comment by Udera** [comment]
I finally used the `+*` suffix. It makes more sense to me (I disabled the `-*`-suffix because of the problems with normal mail addresses.
I uncommented the trusted_user completely.

File: docs/configure [hunk 3]
```diff
@@ -749,8 +752,8 @@ system_aliases:
   allow_fail
   allow_defer
   data = ${lookup{$local_part}lsearch{/etc/aliases}}
-  user = mailnull
-  group = mail
+  user = vexim
+  group = vexim
```
rimas-kudelis: These shouldn't be changed. System aliases have nothing to do with Vexim.

docs/configure
```diff
@@ -478,10 +480,11 @@ ditch_maxmsgsize:
 		and domain = '${quote_mysql:$domain}' \
 		and users.maxmsgsize > 0 \
 		and users.domain_id=domains.domain_id}{${value}K}fail} Kb.
-  local_part_suffix = -*
+  local_part_suffix = +*
```
rimas-kudelis: What about using an `.ifdef  ... .endif` like I suggested in a previous comment? This would make toggling this feature globally much easier.

vexim/adminaliasaddsubmit.php
```diff
@@ -67,7 +67,7 @@
        ':localpart' => $_POST['localpart'],
        ':username' => $_POST['localpart'] . '@' . $_SESSION['domain'],
        ':domain_id' => $_SESSION['domain_id'],
-       ':crypt' => crypt_password($_POST['clear'],$salt),
+       ':crypt' => crypt_password($_POST['clear']),
```
rimas-kudelis: This (and siteaddsubmit.php) should have better been a separate PR. I wold have accepted it right away! :)
In future, please try making your PR's as atomic as possible – this makes the review and accepting process much faster, because good changes like this aren't then blocked by unrelated changes which require further discussion or editing.

Udera:
@rimas-kudelis : Thanks for your comments, I put them into the last commit.

rimas-kudelis:
Thank you!

vpm:
Hello, when the release is scheduled?

rimas-kudelis:
It's not really scheduled for a particular date. But if we keep our current pace, it might happen rather soon.

PR #61, opened 2015-09-17
Add more crypt schemes (sha512/bcrypt)

Udera:
After we removed the 'clear'-password field, now I added support for more advanced hashing for user passwords. Sha512 can be used on Linux, exim can authenticate via courier or dovecot-imap. Bcrypt is only available (by default) on *BSD systems.
Configuration and documentation updated.


Udera:
I rebased my request to remove conflicts.

setup/mysql.sql
```diff
@@ -145,10 +147,25 @@ INSERT INTO `vexim`.`users`
...
+(   '1', 'siteadmin', 'siteadmin', '$6$uR.EiB1dj5rrvwMF$Qh5LgdjOZavKXwhi9IF0Yuzu7qxsG.dLTTB8e./55ZRNfBuZVLnfUSOEXa0oWT6174myO.WYkOy83HYWAKPbK/', '65535', '65535', '', '',  'SiteAdmin', 'site', '1'
 );
 
+-- fix password when using bcrypt (on *BSD only) encrypted password:
+-- UPDATE `vexim`.`users` SET `crypt` = '$2a$10$KKtb78YkexNl4Ik3RymPEObqTsk/ivneEHl/Q5TsDpRBGyYjOl33G'
+--   WHERE `user_id` = '1' LIMIT 1 ;
+
 -- fix password when using DES encrypted password:
--- UPDATE `vexim`.`users` SET `crypt` = '0Apup3ZbF9RPg'
+-- UPDATE `vexim`.`users` SET `crypt` = '2Q3wdB2fa10dE'
```
rimas-kudelis: Hm, did this change? Wasn't it `CHANGE` from the very beginning?

Udera: No. I only created test hashes for each hashing method. I didn't retest the old ones. But you can use the old one if you like it better :D

docs/clients/dovecot.txt
```diff
@@ -45,6 +45,25 @@ Configuring Dovecot to work with Virtual Exim
    Dovecot expects the select to return a set of columns with particular
    names, hence usage of the 'AS' keyword above.
 
-3. There's no step 3! Just restart Dovecot, and it should work.
+3. Exim can’t handle SHA512-crypt schemes for passwords, therefore you can use
+   the dovecot authenticator to login:
+   edit conf.d/10-master.conf
```
rimas-kudelis: This part is unclear. Where did you find this file?

docs/clients/dovecot.txt
```diff
@@ -45,6 +45,25 @@ Configuring Dovecot to work with Virtual Exim
...
+       mode = 0660
+       user = vexim
+       }
+     ...
+     }
+
+   and edit conf.d/10-auth.conf
+     auth_mechanisms = plain login
+
+4. Edit your /etc/exim4/exim4.conf or /etc/exim4/conf.d/auth/01_exim4-config_vexim
```
rimas-kudelis: I think we should rename `01_exim4-config_vexim` to `30_vexim_authenticators`. I overlooked this in @Z3po's initial PR, but this current name doesn't make sense (`exim4-config` is the name of the Debian package).
It might also make sense to put your authenticators into a separate file (in which case the current file could be named `30_vexim_basic_auth` and your authenticators could go into `30_vexim_advanced_auth`, or something like that), but splitting is not necessary.

setup/create_db.pl
```diff
@@ -387,12 +390,17 @@ sub add_postgresveximuser {
...
  	{
 	  if ($dbtype eq "mysql") { $dbh = $mydbh; } 
 	#elsif ($dbtype == "pgsql") { $dbh = $pgdbh; }
-	
+        print "Select an hashing scheme for the user passwords:\n";
+        print "sha512:\tcompatible with Dovecot/Courier\n";
+        print "bcrypt:\tcompatible only in *BSD with Dovecot/Courier (do not use for Linux)\n";
+        print "clear:\tcompatible with Dovecot/Courier, not recommended\n";
+        print "\nPlease select a hashing scheme (default: sha512): ";
+        chomp($hashmethod = <STDIN>);
```
rimas-kudelis: Why not make it a command-line option like other settings?

But anyway, you must have noticed that I am thinking about removing this file altogether, so this bit might just turn out to be irrelevant very soon (unless somebody opposes the removal on proper grounds).

setup/create_db.pl
```diff
@@ -21,6 +21,9 @@
 use strict;
 use DBI();
 use Getopt::Long;
+use Crypt::Eksblowfish::Bcrypt qw(bcrypt);
+use Crypt::OpenSSL::Random;
+use MIME::Base64 ();
```
rimas-kudelis: Do these modules exist in default installations?

docs/clients/dovecot.txt
Udera: This is a dovecot-configuration file. We should probably link the original documentation too
http://wiki2.dovecot.org/HowTo/EximAndDovecotSASL

setup/create_db.pl
Udera: I noticed after my pull request and after I spent some time to get this working in perl ;-( Anyway, it would make more sense to make these installation-scripts and migration steps in php as well (for the future).

setup/create_db.pl
Udera: Speaking of debian this Eksblowfish and Openssl-Libraries for perl can be installed via packet manager. I didn't find a proper way with standard-functions (I can only copy&paste-perl) and you shouldn't write your own crypto. Like this they are working and generating hashes that our vexim-scripts can handle.

vexim/config/functions.php
```diff
@@ -127,6 +127,28 @@ function crypt_password($clear, $salt = '')
                     }
                     $cryptedpass = crypt($clear, $salt);
                 break;
+		case 'sha512':
```
rimas-kudelis: I think I have another "ingenious" idea. :) Why don't we allow the users to specify salt prefix instead of hash type? We could still special-case '{SHA}' and 'clear', but that would make the number of algorithms Vexim accepts out-of-the-box completely dependent on what's available on a particular system in a particular time. E.g. for SHA512, the user could to specify `$salt_prefix = '$6$'`, and for BCRYPT – `$salt_prefix = '$2a$10$' (or whatever other number of iterations they want). The `crypt_password()` function would only have to use a salt string long abd clean enough (e.g. 128 characters) and let crypt() do its work.
Thoughts?

Udera: This would leave the user with more configuration options. But we still have fixed salt-lengths in our code, for SHA512 you can even add more parameters `$6$rounds=5000$salt$` that needs all to be compatible. We would have to test first if all these schemes are compatible with courier and dovecot (even cyrus?).

rimas-kudelis:
I thought we could leave final testing to the admin setting their system up.
What we could do is perhaps list the methods that we have tested and provide good examples and sane defaults.
I tried playing with the crypt() function, and it doesn't seem like a too long salt would break anything, so I assume we could always use the salt long enough to satisfy all commonly used algorithms.

Udera:
I just put in some of your remarks.

Regarding the hash and salting, I don't really see the point for a user to change all these settings by default. I haven't seen any solution where you can change all these parameters. If you think it's better to use a longer salt or to use more rounds, let's do this straight away.

rimas-kudelis:
My idea was that we don't have to limit our users to a certain list of crypt algorithms. If we allow specifying salt prefix directly, we wouldn't have to keep track of what algorithms are available/practical at any given time and what their prefixes are. For example, the [Blowfish-based crypt scheme](https://en.wikipedia.org/wiki/Crypt_%28C%29#Blowfish-based_scheme) may use `$2a$`, `$2b$`, `$2x$` or `$2y$` prefixes, with the latter two introduced only in 2011. I presume even `$2$` was valid at some point. If we allow the admin to specify "blowfish" as crypt scheme, we have to choose only one prefix for that scheme, and assume that it will always be practical, or that the admin will edit Vexim code to cater with the new reality, or that we will release an update and the admin will use it. 10 iterations may be secure enough today, but not so in a few years. I'm just saying we have a choice here – we can allow the user to choose the password encryption algorithm from a list, or we can let them enter it free-form, but make the entry itself a bit more complicated. Or we can even support both ways, shipping our friendly-named defaults, yet allowing custom crypt schemes starting with the `$` character.

Udera:
What I like about your idea is that you can even increase security over time. Assume you start with sha512 and 5000 rounds (default). Later, you can add more rounds in the configuration and all new passwords will respect this new definition. Previous passwords are still valid.
So I try to implement the salt-prefix into the crypt scheme. I would keep the options des, md5 and clear for compatibility.

Udera:
For the moment, I kept sha512 and bcrypt. You can set a prefix like you want, e.g. $6$rounds=N$ or $2y$10$, $2a$12$, ...

docs/clients/dovecot.txt
```diff
@@ -45,6 +45,28 @@ Configuring Dovecot to work with Virtual Exim
...
+     ...
+     }
+
+   and edit /etc/dovecot/conf.d/10-auth.conf
+     auth_mechanisms = plain login
+
+4. Edit your /etc/exim4/exim4.conf or /etc/exim4/conf.d/auth/30_vexim_authenticators
+   and uncomment the dovecot-authenticator section. 
+
+5. There's no step 3! Just restart Dovecot and Exim, and it should work.
```
rimas-kudelis: There's no step 5, not 3. :)
Also, I think you should add yourself to the credits below, or remove them altogether.
Although, if we move our documentation to the Wiki, this file will probably become unnecessary altogether.

docs/debian-conf.d/auth/30_vexim_authenticators
```diff
@@ -0,0 +1,105 @@
+
+### 01_exim4-config_vexim
```
rimas-kudelis: This should contain the new file name.

setup/mysql.sql
```diff
@@ -145,10 +147,25 @@ INSERT INTO `vexim`.`users`
     smtp, pop, realname, type, admin
 )
 VALUES 
-(   '1', 'siteadmin', 'siteadmin', '$1$12345678$2lQK5REWxaFyGz.p/dos3/', '65535', '65535', '', '',  'SiteAdmin', 'site', '1'
+(   '1', 'siteadmin', 'siteadmin', '$6$uR.EiB1dj5rrvwMF$Qh5LgdjOZavKXwhi9IF0Yuzu7qxsG.dLTTB8e./55ZRNfBuZVLnfUSOEXa0oWT6174myO.WYkOy83HYWAKPbK/', '65535', '65535', '', '',  'SiteAdmin', 'site', '1'
```
rimas-kudelis: I wonder, maybe we can leave this password unchanged? The user should change it right after logging in for the first time, so assuming they configure Vexim prior to logging in, their updated password will be hashed more securely anyway.
For same reason, I suggest to only leave CLEAR (and DES and/or SHA, if logging in with these as the configured crypt method would not work) as commented-out alternatives. Or maybe even drop all of these alternatives from our file, since we are discouraging from using these algorithms for auth anyway.

setup/pgsql.sql
```diff
@@ -76,7 +76,7 @@ INSERT INTO users (domain_id, localpart, username, crypt, uid, gid, smtp, pop, r
   		VALUES ('1',
 		'siteadmin',
 		'siteadmin',
-		'\$1\$12345678\$2lQK5REWxaFyGz.p/dos3/',
+		'$6$uR.EiB1dj5rrvwMF$Qh5LgdjOZavKXwhi9IF0Yuzu7qxsG.dLTTB8e./55ZRNfBuZVLnfUSOEXa0oWT6174myO.WYkOy83HYWAKPbK/',
```
rimas-kudelis: same here

setup/mysql.sql
Udera: A MD5-hashed password by default is not a good idea. I thought about using SHA512 as default because it seems to be supported by all systems (Courier/Dovecot on Linux and BSD). I removed the old encryption schemes as they shouldn't be used in new setups. I left clear-text passwords.

Made some changes to the documentation. Moving all the documentation to the wiki should be made in a different pull request.

vexim/config/functions.php
```diff
@@ -127,11 +127,28 @@ function crypt_password($clear, $salt = '')
...
+                    }
+                    $cryptedpass = crypt($clear, $salt);
+		break;
+		case 'bcrypt':
+                    if (empty($salt))
+                    {
+                        $salt = '$2a$10$'.get_random_bytes(22).'$';
+                    }
+                    $cryptedpass = crypt($clear, $salt);
+		break;
```
rimas-kudelis: I'd like to consolidate this function as much as possible.
From my understanding, we don't really have to know the length of the salt each time, since the algorithm in use will simply ignore all unnecessary characters. Nor does the salt have to end with the dollar sign. Thus, I think we could set only the salt prefix in this `switch` block, and only do that when the salt itself is empty. After setting the prefix, we could run `get_random_bytes()` with `CRYPT_SALT_LENGTH` as an argument.
Only exceptions to this scheme would be the algorithms which don't use `crypt()`, namely, `sha` and `clear`.
Also, the default case statement below could have a simpler regex used in `preg_match()` in this case: something like `/^\$[:digit:][:alnum:]?\$/`. I don't think we should enforce only `$2*$` and `$6$` here.

vexim/config/variables.php.example
```diff
@@ -29,10 +29,14 @@
...
 
-  /* Set to either "des" or "md5" depending on your crypt() libraries
-     or "clear" for clear-text passwords.
-     It is not recommended to use the "clear" option */
-  $cryptscheme = "md5";
+  /* Set to either "sha512" or "bcrypt" (only on *BSD) for advanced
+     pw-hash functions, "des" and "md5" (kept for compatibility
+     to older setups), or "clear" for clear-text passwords.
+     It is not recommended to use the "clear" option 
+     You can give more details into the hashing algorithm, e.g.
```
rimas-kudelis: I suggest to rewrite this line like this:
```
Alternatively, you can specify custom salt prefix here, e.g.
```

Udera:
Latest commit https://github.com/Udera/vexim2/commit/c7e2a61c3d50b10c6677b1c76a68a370b30bbb2f
If password "CHANGE" is used the siteadmin is redirected to the change-password-page after login.

setup/mysql.sql
rimas-kudelis: My point was, it's not a good idea to keep the default password in a production system anyway. `$1$` works in all setups, while `$6$` might not (at least to my understanding, correct me if I'm wrong). It doesn't matter which encoding scheme we use in our default sql file, because everyone knows what the password behind that is anyway. I would actually opt for keeping it in plain text, if it worked, but it doesn't. :)

Udera: I think $6$ works on all modern systems (at least if you run your setup with courier or dovecot). If you have a more exotic setup and you are not sure, I would recommend to use clear-text password first because it is easier to switch to any hashing algorithm after. With $1$ that is not possible.
https://github.com/Udera/vexim2/commit/c7e2a61c3d50b10c6677b1c76a68a370b30bbb2f should solve the problem with the default password.

rimas-kudelis: OK, if it works without Courier or Dovecot, it can stay, although I still don't see a point in this, as this password is supposed to be changed ASAP, and the new one will be hashed using the algorithm configured in the settings. But if this only works when authenticating against Courier or Dovecot, I don't think it's OK.

Udera: Cyrus: Uses pam_mysql, good question if this works with sha512.
mysql-popper: uses crypt() of the system
tpop3d: ?
qpop-mysql: ?

Debian Jessie does use SHA512 in /etc/shadow, so I suppose that crypt() and pam_mysql should be possible.

But this does not solve our problem. Even if we set "CHANGE" in cleartext, enforced a password change after the first login (then with the selected hashing scheme), this wouldn't enable any of the POP/IMAP server to be able to speak our crypt scheme. We need to clarify before which crypt-schemes our IMAP/POP-server (and exim in case it isn't courier/dovecot) can handle.

Udera:
I removed some lines, we can even put MD5 into this generalized salt-part.

Udera:
Should be still working ;-)

docs/clients/courierimap.txt
```diff
@@ -9,6 +9,11 @@ Installing Courier-IMAP 2.x on FreeBSD:
       cp imapd.dist imapd
       vi imapd
 
+Installing Courier-IMAP on Debian: 
+      apt-get install courier-imap courier-imap-ssl courier-pop courier-pop-ssl courier-authlib-mysql courier-authdaemon
```
rimas-kudelis: It might be worth mentioning that `courier-pop` and `courier-pop-ssl` are not necessary unless you want to run a POP3 server (just as `courier-imap` and `courier-imap-ssl` are only required if you want to run an IMAP server), although it's probably quite obvious anyway.

docs/clients/dovecot.txt
```diff
@@ -36,7 +36,7 @@ Configuring Dovecot to work with Virtual Exim
 
       driver = mysql
       connect = host=/var/run/mysqld/mysqld.sock dbname=vexim user=vexim password=CHANGE
-      default_pass_scheme = CRYPT
+      default_pass_scheme = SHA512-CRYPT #BLF-CRYPT (<- this is for bcrypt, only in *BSD!) #PLAIN
```
rimas-kudelis: Do you know of a case where simply `CRYPT` would not work with our setup? I couldn't quickly find a source which would list which ciphers are available in which distributions, but I suspect it might be quite uniform, in which case we could just use tell software to use the crypt() implementation provided by the OS.

docs/clients/dovecot.txt
```diff
@@ -45,6 +45,26 @@ Configuring Dovecot to work with Virtual Exim
    Dovecot expects the select to return a set of columns with particular
    names, hence usage of the 'AS' keyword above.
 
-3. There's no step 3! Just restart Dovecot, and it should work.
+3. Exim can’t directly handle SHA512-crypt for user authentication but exim can use
```
rimas-kudelis: Same here. If most current Linux distros support SHA-512 out of the box, we could use that as an excuse not to tell Exim to authenticate against other products by default

docs/configure
```diff
@@ -1025,6 +1025,75 @@ login_virtual_exim:
         {*}{}}
 
 
+# If you use more advanced password hashing (sha512 or bcrypt), exim cannot handle
```
rimas-kudelis: If there are cases where Exim CAN handle these hashes itself, here we should say that it "may not be able to handle" instead of "cannot handle" them.

docs/debian-conf.d/auth/30_vexim_authenticators
```diff
@@ -0,0 +1,104 @@
...
+	    # comment the next lines out if you want to be able to authenticate
+	    # without using SSL or TLS
+        server_advertise_condition = ${if or{\
+           {!eq{$tls_cipher}{}}\
+           {match_ip {$sender_host_address}{@[]}}\
+            }\
+            {*}{}}
+
+
+# If you use more advanced password hashing (sha512 or bcrypt), exim cannot handle
```
rimas-kudelis: Same here.

setup/mysql.sql
rimas-kudelis: Can't tell about all these products, never used them myself...

Regarding Jessie: indeed, if SHA512 is now used in /etc/shadow by default, it probably indicates that this is available in system `crypt()` now.

I don't quite understand your last point though. `$1$` is very likely to be uniformly supported, but even if not, the siteadmin user should not be logging into IMAP or POP3 anyway. For one, his home directories are not specified by default. I wonder if anyone ever tried logging in over IMAP as siteadmin. :)

vexim/config/functions.php
```diff
@@ -99,39 +99,37 @@ function alpha_menu($flag)
...
-		default:
-		   $cryptedpass = crypt($clear, $salt);
+              default:
+                if(preg_match('/\$[:digit:][:alnum:]?\$/', $cryptscheme)) {
+                  $salt=$cryptscheme;
+                } else {
+                  die(_('The value of $cryptscheme is invalid!'));
+                }
+            }
+           $salt.=get_random_bytes(CRYPT_SALT_LENGTH);
```
rimas-kudelis: Huh, i think I know why Avleen appended the `$` here: because of [this](http://man7.org/linux/man-pages/man3/crypt.3.html#NOTES). I wonder if there's a mistake in this manpage, or if this behavior has been changed and this bit of the manual overlooked, or if PHP just works differently... Either way, I guess it might be safer to add it back. I'm sorry for being so inconsistent with my comments.

Also, spacing is still inconsistent in this block.

docs/clients/courierimap.txt
Udera: You won't break anything if you install a POP3 and IMAP4 server.

docs/clients/dovecot.txt
Udera: Yes, you are right this should work. This makes life much easier.
glibc >= 2.7 is needed, so no real restriction.

Udera:
I applied the changes.

docs/clients/courierimap.txt
rimas-kudelis: Of course not, I'm just saying they aren't really required.

setup/mysql.sql
Udera: As we change the password after the first login, the crypt scheme technically doesn't matter. What I thought of, if someone checks out the basics of password hashing and sees that $1$ refers to md5, he might think that if we decide to use this for the super-admin password this will be sufficient for his user's passwords as well.

For compatibility we shouldn't worry too much, it requires php 5.3.2 (http://www.w3schools.com/php/func_string_crypt.asp), which is already EOL (https://secure.php.net/eol.php).

https://www.bytepark.de/blog/2013/10/list-of-current-php-version-in-major-linux-distros/

For the crypt() function, glibc 2.7 or higher is needed, which was widely introduced some time back:
    Debian 5.0 lenny
    Fedora 8 werewolf
    Gentoo 2008.0
    Mandriva 2008.1
    Slackware 12.1 ... 12.2
    Turbolinux 12D magny-cours
    ubuntu 8.04 LTS Hardy Heron

rimas-kudelis: You're right, I failed to consider the psychological impact. :)

Udera:
Ok, they are back in.

rimas-kudelis:
Good!
I haven't looked at the Perl script changes, as I want to remove it before the release.
One last thing: could you now squash all these commits into one, and give that commit the same name as this PR?

Udera:
You mean a new PR?
I could do this tonight.

rimas-kudelis:
No, not necessarily a new PR. You may push to the same branch (with `--force`) after squashing in which case this PR will be updated.
It's not really elegant as it rewrites history, but assuming that your branch hasn't yet been cloned by anyone your branch yet, it's safe.
How to squash really simply is explained [here](http://stackoverflow.com/questions/8213926/git-simplest-way-of-squashing-commits-on-master#8214049).
Note that if you don't feel confident about squashing, you may want to copy your repository before doing that. Or you can ask me to do that for you. :)

Udera:
I changed history :D

rimas-kudelis:
Congrats! :)

PR #64, opened 2015-09-21
add spam_drop for alias-address

Udera:
We introduced the spam_drop-function that allowed the direct deletion of spam mails (instead of moving it to a special spam-directory. We didn't add this feature for alias-addresses (mail forwarding), all mails were forwarded. With this feature you can decide if a spam mail beyond the sa-refuse-tag is either deleted or forwarded.
Forwarded spam-mails can lead to a bad reputation of a mail server.

### DIFF
--- a/vexim/adminaliaschange.php
+++ b/vexim/adminaliaschange.php
@@ -11,7 +11,16 @@
   if ($sth->rowCount()) {
     $row = $sth->fetch();
   }
+  $domquery = "SELECT avscan,spamassassin,quotas,pipe FROM domains
+    WHERE domain_id=:domain_id";
+  $domsth = $dbh->prepare($domquery);
+  $domsth->execute(array(':domain_id'=>$_SESSION['domain_id']));
+  if ($domsth->rowCount()) {
+    $domrow = $domsth->fetch();
+  }
 ?>
+
+
 <html>
   <head>
     <title><?php echo _('Virtual Exim') . ': ' . _('Manage Users'); ?></title>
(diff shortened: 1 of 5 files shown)


vexim/adminaliaschange.php
```diff
@@ -99,6 +108,9 @@
               } ?> class="textfield">
             </td>
           </tr>
+          <?php
+           if ($domrow['avscan'] == "1") {
+        ?>
```
rimas-kudelis: I wonder if this check (and the one checking if SA is enabled) is really necessary. It might be a good idea to set these values to something useful even when anti-virus or SA checks are disabled for a particular domain, don't you think so?

Udera: What do you mean, to disable the field but keeping it visible? Or to enforce a default spam/antivirus-check?

rimas-kudelis:
I thought perhaps to leave these options available for editing, even if don't end up being used. But I'm not sure myself if it's necessary or not. My concern here is mostly about what happens when SA or AV is enabled for a domain, for which it was previously disabled.

Udera:
For mailboxes, you don't see any options that you cannot change. For alias-addresses you can change SA or AV settings but they are not written to the database.

rimas-kudelis:
I see. Thanks!

rimas-kudelis:
Is this already implemented in our exim config files?

Udera:
It was introduced to the conf.d-files some time ago:
https://github.com/avleen/vexim2/blame/master/docs/debian-conf.d/router/249_vexim_ditch_routers#L44-L56
I already adjusted the configure-file (https://github.com/avleen/vexim2/commit/b5ac8383bdca4b68257be3eee82cab752cf13f50) and then tried to make this feature accessible via the vexim interface.

PR #67, opened 2015-09-24
Fix to Disable AV and SA for all users in domain

Kami-no:
Fixes bug that reset all settings for all users in domain.


vexim/sitechangesubmit.php
```diff
@@ -4,6 +4,9 @@
   include_once dirname(__FILE__) . "/config/functions.php";
   include_once dirname(__FILE__) . "/config/httpheaders.php";
 
+  $success = FALSE;
```
rimas-kudelis: Please use lowecase here.

PR #68, opened 2015-09-28
updated to some recent changes

Udera:
README.md


README.md
```diff
@@ -233,6 +233,8 @@ Mailman needs to be installed if you want to use mailing lists. The default loca
 
 #### Mail storage and Delivery:
 The mysql configuration assumes that mail will be stored in /var/vmail/domain.com/username/Maildir. If you want to change the path from '/var/vmail/', you need to edit the file:
+>vexim/config/variables.php.example
```
rimas-kudelis: Please review this whole chapter. The file should now always be copied, so this needs to be updated as well. 

Also, please add empty lines around the ones with file names, as it appears otherwise the three lines are formatted as one.

Udera:
The copying was already in the file. I changed a section about the pw-hash and status of the config-file.

PR #72, opened 2015-09-30
Prepare data migration scripts for 2.3. Fixes #59

### LINKED ISSUE #59: Prepare data migration scripts for 2.3
Our database structure has been changing over the years, and it looks like we never really had proper migration scripts.

I think we should provide them in 2.3 to make upgrades easier to users of older releases.

rimas-kudelis:
This adds incremental MySQL migration scripts for those upgrading from older releases of Vexim. The scripts must be run sequentially to upgrade the database step-by-step.


Udera:
I added a small part for users with Schirmacher-patch: https://github.com/Udera/vexim2/commit/b463e007830680934597a297a3e1f533a2133291

rimas-kudelis:
If I remember correctly, my scripts are still somewhat buggy, but I don't remember, what exactly is wrong. :)

Do you think migration from Arne's schema should be in that file, and not in a separate one?

Udera:
Could be a separate one as well but then you need to respect the order. Or do you mean a completely independent `vexim_schirmacher_to_2.3.sql`?

rimas-kudelis:
No, I rather meant something that would apply on top of `vexim_2.2_to_2.3.sql`, but hadn't really though about a name yet. Indeed, order matters here, so if we don't come up with a good one, we might just leave these lines commented out in the main file like you suggested.

Udera:
Don't forget to change the `user`-table:
`localpart`=192 (should be 64) and `domain`=64 (should be 255) 
ref: https://github.com/vexim/vexim2/issues/108

rimas-kudelis:
Before continuing this effort, I think we should first agree to make no more DB changes for this release, otherwise it might be a bit annoying to keep updating the update scripts. :)

rimas-kudelis:
Although it doesn't seem like we have any pending PR's or issues related to a DB, so I might just go on and do my work...

Udera:
Except for the `domain` and `localpart`, there is nothing planned so far.

Udera:
Just a little feedback from my side: I migrated an old 2.2-version with several manual modifications (what we did recently). I ran the upgrade script on the database, however some indexes and stuff didn't work out a 100%. So, I created a new database. Then I dumped only the data from the old and converted and inserted it into the new one. So I've a proper new database with all my old data ;-)

rimas-kudelis:
Sorry, I'm not quite following you. :)

Udera:
apply `vexim_2.2_to_2.3.sql` -> old_db -> export (only data)
`vexim2/mysql.sql` -> new_db -> import data from old_db

Now, I'm using the new_db

rimas-kudelis:
Hm, it's probably due to the fact that I haven't updated my scripts in a while.

rimas-kudelis:
This is now ready for review. Should we decide to add further changes to the schema, these should be discussed or mentioned here.
I'm not sure what to do with the PostgreSQL setup script and migrations: the former is now probably somewhat outdated, and the latter are completely lacking.

Udera:
I just installed vexim2.2.2 (https://www.freshports.org/mail/vexim/). It seems there have been some changes. The default configuration has a few minor issues, the `headers_remove` and missing SSL. Webinterface works right away and can be used to create accounts. I used it for one domain, an alias, created a few account and forwarding addresses (didn't have time to go all this spam/clam-av-stuff).

The migration script ran without errors. To make the code run, I needed to install php-pdo and php-openssl and I could directly login. Per default it still uses MD5-hashes. Unless you change the password, it keeps the `MD5`-scheme:
````
+------------+------------------------------------------------------------------------------------------------------------+
| localpart  | crypt                                                                                                      |
+------------+------------------------------------------------------------------------------------------------------------+
| siteadmin  | $6$hpyGkCxuRYPMhKVD$W4NR3GeXuzZLHXzQM0eWrWPe4f9.qcRIMST8mLHRC4Kqw0WP99sd0U07LdeYXa7KfYIAPtIBN9rhTda4GWjau1 |
| postmaster | $1$OWUUdgzP$sYxS.2I7n1jfgjpBbqyT1/                                                                         |
| uu         | $1$iyw1b0R5$Ok00lHLY.slg/bVRXzNnG.                                                                         |
| kk         | $1$ai4p9lue$85cpWw/FgENLYFst.4acL1                                                                         |
| ss         | $1$tzIkCzPf$fhGIjxqazk8rCa.KStEWx/                                                                         |
| mm         | $1$EJDRyLH1$oeGubwV9k.wycepFkVUPh1                                                                         |
+------------+------------------------------------------------------------------------------------------------------------+
````

What do we want to do about it? Options:

* do nothing and hope the users will change their pw one day
* if the pw-hash is not the one in `variables.php`,  update the hash-scheme
* if the pw-hash is not the one in `variables.php`,  ask the user to change their password (like for the "CHANGE"-password

**Comment by rimas-kudelis** [comment]
OK, I'll add the statement devised in #59 to the last SQL file before dropping the `clear` field:
```
UPDATE `users` 
    SET `crypt`=ENCRYPT(`clear`, CONCAT('$6$', MD5(RAND()))) 
    WHERE `crypt`=ENCRYPT(`clear`, `crypt`);
```
We probably want to define '$6$' as an SQL variable at the top of the file though, to make it easier to change. Or do we?

I think it's also possible to add an EXIT statement at the top of the file to force the user to edit it before applying, although it hardly seems worthwhile.

**Comment by rimas-kudelis** [comment]
Done, check this out.

File: setup/migrations/vexim_2.2_to_2.3_mysql.sql [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+-- passwords match using a new, more secure hashing scheme (SHA-512 by default).
+-- Below you may change this prefix if you prefer a different hashing scheme, e.g. '$2a$10$' for bcrypt
+-- (*BSD-only Blowfish hashing scheme).
+SET @NEW_PW_PREFIX='$6$';
+-- Comment out this definition if you do not want your password hashes updated. Be advised that it will
+-- leave these passwords less secure for no good reason.
+-- In case you want to store passwords in unencrypted form (PLEASE DO NOT DO THIS), uncomment the following
+-- UPDATE statement that follows instead. Note this will cut very long passwords to 48 characters. If this
+-- is a problem for you, move the uncommented statement just above the statement that re-hashes passwords.
+-- UPDATE `users` SET `crypt` = `clear`;
```
rimas-kudelis: I would be very happy to drop these four lines from this file. I think they should either live in the wiki, or even be relayed via mailing list only. :)

Udera: Yes, we can shorten this. `sha512` should be doable by everybody with a current OS and we should not promote clear-text passwords at all. Perhaps we should leave a short note for people who manually deactivated the clear-text option, but that could go into a release-note or changelog as well.

rimas-kudelis:
PR updated. Please review and merge if you think it's fine.

runout-at:
the users.forward field should be much longer too.
255 -> 4096 ?
how many addresses should go in here?

rimas-kudelis:
@Udera, do you want to merge this in or not?

Udera:
Thanks for reminding me. I wanted to test the changes with my current 
test setup first. Will see, if I find some time tonight or tomorrow.

PR #74, opened 2015-10-01
Add PW-generator to siteadd-page

Udera:
Adding the pw-generator to the siteadd-page.


rimas-kudelis:
Huh, what's the point of having `colspan="2"` in the same cell of all rows of a table? I see I've added these years ago when implementing this feature for admins, but it was actually by mistake, they're a remnant of my experimentation with the layout.

Udera:
The table has three columns: https://github.com/Udera/vexim2/blob/pw_gen/vexim/siteadd.php#L24-L28
Look at the whole file not only the changes.

rimas-kudelis:
Oh, I see. Sorry!

PR #75, opened 2015-10-02
Fail if domain mail directory does not exist when creating a new domain. Fixes #71.

### LINKED ISSUE #71: A domain should not be created if its home dir cannot be created
If "Domain Mail directory" is filled incorrectly when creating a new domain (e.g. I had /var/vmail there, but this dir did not exist, and Apache didn't have permission to create it), Vexim will set new domain's homedir to `/domain.name/` (instead of `/var/vmail/domain.name/`). This is wrong and the domain should not be created at all in such situation. Instead, a proper error message should be presented to the user.

rimas-kudelis:
We will also not substitute paths with the result of `realpath()` anymore, and will not accept paths not starting with a `/` character.
The check for actual existing of the directory may be disabled in settings, if that is necessary.

### DIFF
--- a/vexim/siteaddsubmit.php
+++ b/vexim/siteaddsubmit.php
@@ -62,10 +62,21 @@
     die;
   }
   if(isset($_POST['maildir']) && isset($_POST['localpart'])) {
-    $smtphomepath = realpath($_POST['maildir'] . "/") .
-      "/" . $_POST['domain'] . "/" . $_POST['localpart'] . "/Maildir";
-    $pophomepath = realpath($_POST['maildir'] . "/") .
-      "/" . $_POST['domain'] . "/" . $_POST['localpart'];
+    if (substr($_POST['maildir'], 0, 1) !== '/') {
+      header ("Location: site.php?failmaildirnonabsolute={$_POST['maildir']}");
+	  die();
+    }
+    if ($testmailroot && is_dir(realpath($_POST['maildir'])) === false) {
+      header ("Location: site.php?failmaildirmissing={$_POST['maildir']}");
+      die();
+	}
+    $domainpath = $_POST['maildir'];
+    if (substr($domainpath, -1) !== '/') {
+      $domainpath .= '/';
+    }
+    $domainpath .= $_POST['domain'];
+    $smtphomepath = $domainpath . "/" . $_POST['localpart'] . "/Maildir";
+    $pophomepath = $domainpath . "/" . $_POST['localpart'];
   }
 //Gah. Transactions!! -- GCBirzan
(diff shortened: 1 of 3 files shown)


vexim/siteaddsubmit.php
```diff
@@ -62,10 +62,21 @@
...
   if(isset($_POST['maildir']) && isset($_POST['localpart'])) {
-    $smtphomepath = realpath($_POST['maildir'] . "/") .
-      "/" . $_POST['domain'] . "/" . $_POST['localpart'] . "/Maildir";
-    $pophomepath = realpath($_POST['maildir'] . "/") .
-      "/" . $_POST['domain'] . "/" . $_POST['localpart'];
+    if (substr($_POST['maildir'], 0, 1) !== '/') {
+      header ("Location: site.php?failmaildirnonabsolute={$_POST['maildir']}");
+	  die();
+    }
+    if ($testmailroot && is_dir(realpath($_POST['maildir'])) === false) {
```
Udera: If $testmailroot==0, you don't want to check the realpath. But if $testmailroot==0 this line will always be false because 
`FALSE && FALSE === FALSE`, `FALSE && TRUE === FALSE`. It should be
````
if ($testmailroot===1) { 
   if (is_dir(realpath($_POST['maildir']))===false) {
      ...
````
Or shorter:
````
if(!$testmailroot || is_dir(realpath($_POST['maildir'])) === false)
````
or:
````
if($testmailroot && !is_dir(realpath($_POST['maildir'])) === true)
````

rimas-kudelis: I think you misunderstood the code, or my intentions.

First of all,
`if ($testmailroot && is_dir(realpath($_POST['maildir'])) === false)`
is equivalent to
`if ($testmailroot && (is_dir(realpath($_POST['maildir'])) === false))`
and is more or less the same as your last suggestion.

Second, if I remember correctly, `is_dir()` will not even be run if `$testmailroot` is `false`, because in that case the result of `&&` is already known to be `false`. So yeah, `realpath()` shouldn't be executed if the setting is false.

Now, your second example is actually incorrect: I do not want to trigger `header()` and `die()` if `$testmailroot` is `false`, regardless of whether `realpath()` fails or succeeds. That's the whole point of that setting! :)

Udera: I misunderstood the line and read it like this:
`if (($testmailroot && is_dir(realpath($_POST['maildir']))) === false)`
which would make no sense.

Only reading problems, I already started to doubt myself :relieved:

Udera:
Checked your code and works as it is supposed to. :+1:

rimas-kudelis:
Thanks for testing! :)

PR #77, opened 2015-10-06
Verify forward e-mail address

Udera:
In the forms it is said, that the forward address must be a valid e-mail address. Actually this was never checked.

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -53,11 +53,15 @@
     header("Location: adminalias.php?badname={$_POST['localpart']}");
     die;
   }
-  if(!filter_var($_POST['smtp'], FILTER_VALIDATE_EMAIL)) {
-    header ("Location: adminalias.php?invalidforward=".htmlentities($_POST['smtp']));
-    die;
+  $forwardto=explode(",",$_POST['smtp']);
+  for($i=0; $i<count($forwardto); $i++){
+    $forwardto[$i]=trim($forwardto[$i]);
+    if(!filter_var($forwardto[$i], FILTER_VALIDATE_EMAIL)) {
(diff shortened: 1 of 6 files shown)


vexim/adminaliasaddsubmit.php
```diff
@@ -53,7 +53,10 @@
     header("Location: adminalias.php?badname={$_POST['localpart']}");
     die;
   }
-
+  if(!filter_var($_POST['smtp'], FILTER_VALIDATE_EMAIL)) {
+    header ("Location: adminalias.php?invalidforward=".htmlentities($_POST['smtp']));
+    die;
+  }
   $aliasto = preg_replace("/[', ']+/", ",", $_POST['smtp']);
```
rimas-kudelis: As you can see here and in `adminaliaschangesubmit.php`, we seem to support multiple comma-separated forward addresses. `filter_var()` would fail these.

rimas-kudelis:
If you really want validation, I think you'll have to `explode()` these variables using comma as delimiter, `trim()` and `filter_var()` all elements of the resulting array, and then `implode()` the array again. Or something similar.

Udera:
Good to know. I updated the scripts for alias-addresses.

rimas-kudelis:
Don't you think `array_walk()` would be nicer?

Udera:
For `array_walk()`, I would need to create a function (perhaps even it into `config/functions.php`). I don't know if it makes the code more readable. Then we could as well merge the `adminaliasaddsubmit.php` and `adminaliaschangesubmit.php` that will remove much more identical code.

I first tried a `foreach()`-loop but the value inside the loop is only loop-local (`trim($element)` has no effect on the array outside of `foreach()`). I could have passed pointers but finally I just tried a solution that is working.

rimas-kudelis:
Fair enough. :)

PR #79, opened 2015-10-07
Password check operator

Udera:
Using the !== operator on the password. With hashed passwords, this is no problem, but if clear-passwords are used this could be a problem. ("01"=="1").

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -31,7 +31,7 @@
     $_POST['on_spamassassin'] = 0;
   }
   # If a password wasn't specified, create a randomised 128bit password
-  if (($_POST['clear'] == "") && ($_POST['vclear'] == "")) {
+  if (($_POST['clear'] === "") && ($_POST['vclear'] === "")) {
     $junk = md5(rand().time().rand());
     $_POST['clear'] = $junk;
     $_POST['vclear'] = $junk;
@@ -62,7 +62,7 @@
     }
   }
   $aliasto = implode(",",$forwardto);
-  if (alias_validate_password($_POST['clear'], $_POST['vclear'])) {
+  if (validate_password($_POST['clear'], $_POST['vclear'])) {
     $query = "INSERT INTO users
       (localpart, username, domain_id, crypt, smtp, pop, uid, gid, realname, type, admin, on_avscan, 
        on_spamassassin, sa_tag, sa_refuse, spam_drop, enabled)
--- a/vexim/adminaliaschangesubmit.php
+++ b/vexim/adminaliaschangesubmit.php
@@ -43,7 +43,7 @@
   }
 
   # Update the password, if the password was given
-  if(isset($_POST['password']) && $_POST['password']!='' ){
+  if(isset($_POST['password']) && $_POST['password']!=='' ){
 	if (validate_password($_POST['password'], $_POST['vpassword'])) {
 		$cryptedpassword = crypt_password($_POST['password']);
 		$query = "UPDATE users SET crypt=:crypt WHERE user_id=:user_id AND domain_id=:domain_id AND type='alias'";
--- a/vexim/adminuserchangesubmit.php
+++ b/vexim/adminuserchangesubmit.php
@@ -130,25 +130,27 @@
   }
 
   # Update the password, if the password was given
-  if (validate_password($_POST['clear'], $_POST['vclear'])) {
-    $cryptedpassword = crypt_password($_POST['clear']);
-    $query = "UPDATE users
-      SET crypt=:crypt WHERE localpart=:localpart
-      AND domain_id=:domain_id";
-    $sth = $dbh->prepare($query);
-    $success = $sth->execute(array(':crypt'=>$cryptedpassword,
-        ':localpart'=>$_POST['localpart'], ':domain_id'=>$_SESSION['domain_id']));
-    if ($success) {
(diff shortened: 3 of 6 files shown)


rimas-kudelis:
We have `validate_password()` and `alias_validate_password()` methods in `functions.php`. Could you adjust these as well?
In addition to checking for equivalency in them, I think we should remove checks for bad SQL characters, because SQL security shouldn't be handled that way, at least in my opinion.
Also, `adminuserchange.php` should call `validate_password()`, since we have that function anyway.

Udera:
`adminuserchange.php` looks good for me. `validate_password()` is already used.
What do you mean with the SQL security? These `if(isset($_POST['spamassassin'])) { $_POST['spamassassin']=1; }` ?

rimas-kudelis:
I meant `adminuserchangesubmit.php`, sorry for confusion. Basically, wherever we have code like this:
```
if (($_POST['clear'] === "") && ($_POST['vclear'] === "")) {
```
we should call `validate_password()` instead (or `validate_alias_password()`).

And where I talked about SQL security, I guess I was daydreaming or something, cause now I can't find the code I was talking about (it disallowed backtick, apostrophe and quote characters in the password). Just ignore that remark of mine.

Udera:
There was one more in `adminaliasaddsubmit.php`

vexim/adminaliasaddsubmit.php
```diff
@@ -31,7 +31,7 @@
     $_POST['on_spamassassin'] = 0;
   }
   # If a password wasn't specified, create a randomised 128bit password
-  if (($_POST['clear'] == "") && ($_POST['vclear'] == "")) {
+  if(!validate_password($_POST['clear'],$_POST['vclear'])) {
```
rimas-kudelis: Please add a whitespace after `if` and after `,`. :)

vexim/adminuserchangesubmit.php
```diff
@@ -146,7 +146,7 @@
       header ("Location: adminuser.php?failupdated={$_POST['localpart']}");
       die;
     }
-  } else if ($_POST['clear'] != $_POST['vclear']) {
+  } else if (!alias_validate_password($_POST['clear'],$_POST['vclear'])) {
```
rimas-kudelis: Also here.
Then please squash these commits into one, and we're good for merge!

Udera:
done.

Udera:
Added the last spacing and cleaned up.

vexim/adminaliasaddsubmit.php
```diff
@@ -31,7 +31,7 @@
     $_POST['on_spamassassin'] = 0;
   }
   # If a password wasn't specified, create a randomised 128bit password
-  if (($_POST['clear'] == "") && ($_POST['vclear'] == "")) {
+  if (!validate_password($_POST['clear'], $_POST['vclear'])) {
```
rimas-kudelis: No wait, this is wrong. It will create a random password even if the fields were non-empty, but didn't match. I'm sure this wasn't intended.
Sorry, I must have overlooked the context of this line.

vexim/adminaliaschangesubmit.php
```diff
@@ -43,7 +43,7 @@
   }
 
   # Update the password, if the password was given
-  if(isset($_POST['password']) && $_POST['password']!='' ){
+  if (isset($_POST['password'])) {
```
rimas-kudelis: According to the comment above, here `$_POST['password']` could be set to an empty value, which indicates that it shouldn't be changed.
I think we should check for existence and non-emptiness of both `$_POST['password']` and `$_POST['vpassword']` here, and only proceed if all four conditions are satisfied.

rimas-kudelis: Actually, no, this should remain as it was, except the comparison operator.

vexim/adminuserchangesubmit.php
```diff
@@ -146,7 +146,7 @@
       header ("Location: adminuser.php?failupdated={$_POST['localpart']}");
       die;
     }
-  } else if ($_POST['clear'] != $_POST['vclear']) {
+  } else if (!alias_validate_password($_POST['clear'], $_POST['vclear'])) {
```
rimas-kudelis: A call to an `alias_something` function here looks a bit dumb, don't you think? I think it would be better to just enclose this `if..else` block in a parent `if`, similar to what's being done in `adminaliaschangesubmit.php`.

rimas-kudelis: Then, this `else` wouldn't need a condition.

vexim/config/functions.php
```diff
@@ -28,7 +28,7 @@ function validate_password($clear,$vclear)
      */
     function alias_validate_password($clear,$vclear) 
     {
-        return ($clear == $vclear);
+        return ($clear === $vclear);
```
rimas-kudelis: I understand I didn't mention this before, but could we perhaps merge these functions into one, with a third param indicating whether or not empty password is acceptable (`false` by default)?

rimas-kudelis:
Please tell me if I pissed you off by these comments. I feel a bit guilty now, so if you want, I can do all these things myself. :)

Udera:
I tried to put in all of your remarks. I found an outdated message, in previous versions some special characters were not permitted. Now this isn't a problem any more (we fixed this last year).

rimas-kudelis:
Thank you!

PR #82, opened 2015-10-11
fix headers_remove in the transport section for exim >4.82. Fixes #81

### LINKED ISSUE #81: Broken exim4-config on debian jessie
I tried to set up a new mail server on debian jessie. Unfortunately, there is still the string expansion problem in the headers_remove statements, like reported here:
http://www.gossamer-threads.com/lists/exim/users/100446

ref in our code:
https://github.com/vexim/vexim2/blob/master/docs/configure#L665-L676

related bug: https://bugs.exim.org/show_bug.cgi?id=1533
but this fix doesn't seem to be in the debian-jessie packages (neither in the testing).

So it makes vexim unusable for debian jessie at the moment.

Udera:
fixes: https://github.com/vexim/vexim2/issues/81

Works on debian jessie. Compatible with exim 4.82 (ubuntu 14.04)

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -647,6 +647,7 @@ virtual_forward:
 
 virtual_domains:
   driver = redirect
+  domains = +local_domains
   address_data = ${lookup mysql{\
         select smtp, users.sa_tag*10 AS sa_tag, users.on_spamassassin AS on_spamassassin, \
         users.uid AS uid, users.gid AS gid, quota from users,domains \
--- a/docs/configure
+++ b/docs/configure
@@ -647,33 +647,32 @@ virtual_forward:
 
 virtual_domains:
   driver = redirect
+  address_data = ${lookup mysql{\
+        select smtp, users.sa_tag*10 AS sa_tag, users.on_spamassassin AS on_spamassassin \
+        from users,domains \
+        where localpart = '${quote_mysql:$local_part}' \
+                and domain = '${quote_mysql:$domain}' \
+                and domains.enabled = '1' \
+                and users.enabled = '1' \
+                and users.domain_id = domains.domain_id}{$value}fail}
   allow_fail
-  data = ${lookup mysql{select smtp from users,domains \
-  		where localpart = '${quote_mysql:$local_part}' \
-		and domain = '${quote_mysql:$domain}' \
-		and domains.enabled = '1' \
-		and users.enabled = '1' \
-		and users.domain_id = domains.domain_id}}
-  headers_add = ${if >{$spam_score_int}{${lookup mysql{select users.sa_tag * 10 from users,domains \
-  		where localpart = '${quote_mysql:$local_part}' \
-		and domain = '${quote_mysql:$domain}' \
-		and users.on_spamassassin = '1' \
-		and users.domain_id=domains.domain_id }{$value}fail}} {X-Spam-Flag: YES\n}{} }
-  # You may have to use single colons in headers_remove rule, or comment it out altogether
-  # if using Exim v. 4.85 or older. See https://github.com/avleen/vexim2/issues/21 and
-  # check out the links pasted in that issue for more info.
+  data = ${extract{smtp}{$address_data}}
+  headers_add = ${if and { \
+                    {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
+                    {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
+                    } {X-Spam-Flag: YES\n}{} }
   headers_remove = ${if or { { <{$spam_score_int}{1} } \
-  			     { <{$spam_score_int}{${lookup mysql{select users.sa_tag * 10 from users,domains \
-			       where localpart = '${quote_mysql::$local_part}' \
-			       and domain = '${quote_mysql::$domain}' \
-			       and users.on_spamassassin = 1 \
-			       and users.domain_id=domains.domain_id}{$value}fail}} } \
-			     { eq {0}{${lookup mysql{select users.sa_tag * 10 from users,domains \
-			       where localpart = '${quote_mysql::$local_part}' \
-			       and domain = '${quote_mysql::$domain}' \
-			       and users.on_spamassassin = 0 \
-			       and users.domain_id=domains.domain_id}{$value}fail}}} \
-			   } {X-Spam-Score:X-Spam-Report} }
+                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                  } \
+                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                          }  {X-Spam-Score}}
+  headers_remove = ${if or { { <{$spam_score_int}{1} } \
+                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                  } \
+                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                          }  {X-Spam-Report}}
   # using local_part_suffixes enables possibility to use user-"something" localparts
   # which could cause you trouble if you're creating email-adresses with dashes in between.
   .ifdef VEXIM_LOCALPART_SUFFIX
@@ -684,7 +683,7 @@ virtual_domains:
   file_transport = virtual_delivery
   reply_transport = address_reply
   pipe_transport = address_pipe
-  
+
 # A group is a list of users
 #
 # if a group is flaged public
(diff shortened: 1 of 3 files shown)


Udera:
I tested on debian jessie with `Exim version 4.84 #2 built 17-Feb-2015 17:01:49`

docs/configure
```diff
@@ -647,33 +647,32 @@ virtual_forward:
...
+  data = ${extract{smtp}{$address_data}}
+  headers_add = ${if and { \
+                    {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
+                    {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
+                    } {X-Spam-Flag: YES\n}{} }
+  headers_remove = ${if or { { <{$spam_score_int}{1} } \
+                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                  } \
+                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
```
rimas-kudelis: Do I understand it correctly that you want to remove `X-Spam-Score` if `on_spamassassin` is `0`? Why? Was it definitely us who added that header?

docs/configure
```diff
@@ -647,33 +647,32 @@ virtual_forward:
...
-			     { eq {0}{${lookup mysql{select users.sa_tag * 10 from users,domains \
-			       where localpart = '${quote_mysql::$local_part}' \
-			       and domain = '${quote_mysql::$domain}' \
-			       and users.on_spamassassin = 0 \
-			       and users.domain_id=domains.domain_id}{$value}fail}}} \
-			   } {X-Spam-Score:X-Spam-Report} }
+                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                  } \
+                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
```
rimas-kudelis: Same here.

docs/configure
Udera: I have only rewritten the rules we had before. But yes, we remove these 
headers, if the user turns off spamassassin. Headers from the sender's 
mail server are probably deleted as well. I have seen that some 
organizations change their spam-headers to `X-Spam-Organization:` that 
they won't be deleted by following mail servers.

Am 2015-10-31 14:54, schrieb Rimas Kudelis:

rimas-kudelis: Oh, I see now.
I wonder if we really want that, it seems logical for me not to touch these headers if SA is disabled altogether for the account.

Udera: You could turn this argument around. Why accepting mail in the first 
place when the sending server already considers this mail as spam?
Problem is, if you leave them in, and you filter your mails with sieve 
or client side (not with the exim-spam-router), anyone could add to his 
mail-header a `X-Spam-Score: -3`, and your filter will accept this 
message.

rimas-kudelis: That's correct, but I doubt you would want to filter your email by spam headers if you disable spam processing for your account. Either you care about these haders and you ask the server to add them, or you just don't care about them.
Then also, `X-Spam-Status` could as well be set to `NO` by the other server, and `X-Spam-Score` could be set to `1` or less. These headers seem quite useless when passed by the other server, but I'm not sure if that means we should remove them.

On the other hand, since it was this way before the patch, I'll merge it in anyway, and we can leave this discussion for later.

docs/debian-conf.d/router/250_vexim_virtual_domains
```diff
@@ -40,34 +40,32 @@ virtual_forward:
 
 virtual_domains:
   driver = redirect
-  domains = +local_domains
```
rimas-kudelis: Is this line really unnecessary?

Udera:
This `domains = +local_domains` was not in the configure file and was probably lost when I made a copy&paste from there. It is probably not necessary because the `data =` will be empty for domains that are not in `local_domains` so the router won't be executed. Though it could be faster to use the `domains=`-selector.

I will fix this and create a new PR.

Udera:
I will look through other routers in a different PR.

rimas-kudelis:
OK, so, I'm not sure that removing `domains = +local_domains` is the right thing to do. Perhaps instead it should be added where missing?

Also, this optimization with `address_data` is really beautiful, perhaps you could apply it to other routers as well? It could be a separate PR, or a part of this one, that's not important. Running a separate SQL query for each field is just way too lame compared to selecting all necessary fields into `address_data` once.

Udera:
I put `domains = +local_domains` back and added it to the `configure`-file as well.

docs/configure
```diff
@@ -867,22 +867,10 @@ virtual_delivery:
...
-		and users.domain_id = domains.domain_id}}
-  group = ${lookup mysql{select users.gid from users,domains \
-		where localpart = '${quote_mysql:$local_part}' \
-		and domain = '${quote_mysql:$domain}' \
-		and users.domain_id = domains.domain_id}}
-  quota = ${lookup mysql{select users.quota from users,domains \
-  		where localpart = '${quote_mysql:$local_part}' \
-		and domain = '${quote_mysql:$domain}' \
-		and users.domain_id = domains.domain_id}{${value}M}}
+  directory = ${extract{smtp}{$address_data}}
```
rimas-kudelis: Are you sure we don't need to define `$address_data` for each router? I think we should, at the very least it would make things clearer, don't you think?

docs/debian-conf.d/transport/30_vexim_virtual_delivery
```diff
@@ -9,22 +9,10 @@ virtual_delivery:
...
-                and users.domain_id = domains.domain_id}}
-  group = ${lookup mysql{select users.gid from users,domains \
-                where localpart = '${quote_mysql:$local_part}' \
-                and domain = '${quote_mysql:$domain}' \
-                and users.domain_id = domains.domain_id}}
-  quota = ${lookup mysql{select users.quota from users,domains \
-                where localpart = '${quote_mysql:$local_part}' \
-                and domain = '${quote_mysql:$domain}' \
-                and users.domain_id = domains.domain_id}{${value}M}}
+  directory = ${extract{smtp}{$address_data}}
```
rimas-kudelis: Same here.

docs/configure
Udera: No, we don't need it: https://books.google.fr/books?id=foCRVaMeRMgC&pg=PA106&lpg=PA106&dq=exim+address_data&source=bl&ots=wrrhBdPqOU&sig=grP98HExo1dLJjQHxROB9dnNK-0&hl=de&sa=X&ved=0CDQQ6AEwAmoVChMIs4zl7e_vyAIVQewUCh364gkZ#v=onepage&q=exim%20address_data&f=false

Like this, we reduce the number of lookups and get a compact, readable code. I don't see the point why we should run the same query twice.

rimas-kudelis: Oh, I see, these are transports, not routers.
Okay, merging then.

Udera:
Reduce more code with the `address_data`-variable is challenging:
I looked a into this. I tried to use one `address_data`-variable for these two routers:
https://github.com/vexim/vexim2/blob/master/docs/configure#L509-L544
(you can preserve data over several routers).
> Because the `address_data` option is not expanded until after a router's preconditions have been tested, you cannot use the value of $address_data in the preconditions of the same router

https://books.google.fr/books?id=foCRVaMeRMgC&pg=PA106&lpg=PA106&dq=exim+address_data&source=bl&ots=wrrhBdPqOU&sig=grP98HExo1dLJjQHxROB9dnNK-0&hl=de&sa=X&ved=0CDQQ6AEwAmoVChMIs4zl7e_vyAIVQewUCh364gkZ#v=onepage&q=exim%20address_data&f=false

If we still want to use the `address_data`, we would need to put an empty router before (which does the database-query to fill `address_data` and nothing else).
````
collection_router:
  data =
  address_data = ${lookup mysql{\
       select ...

spam_drop_router:
...
````

It doesn't look so much nicer because the conditions are more complicated (in the queries, we already put some conditions, like `and spam_drop='1'`, so the `ditch_spam`-router would look like this (without the collection_router):
````
ditch_spam:
    driver = redirect
    allow_fail
    file_transport = virtual_ditch_spam_transport
    data = ${extract{smtp}{$address_data}}
    condition = ${if and { \
                     {eq{0}{${extract{spam_drop}{$address_data}}}} \
                     {eq{0}{${extract{on_forward}{$address_data}}}} \
                     {eq{local}{${extract{type}{$address_data}}}} \
                   }{yes}{no}}
    retry_use_local_part
````
Instead of: https://github.com/vexim/vexim2/blob/master/docs/configure#L523-L544

rimas-kudelis:
Hmm, having a pseudo-router to fill `$address_data` once is both ugly and elegant at the same time.

When I started writing this, I thought it is more ugly than elegant, because it makes an assumption that no other routers will touch this variable.

But now I think that perhaps it's elegant enough, because we **do** list our routers in a particular order, which is important, so prepending all of them with a pseudo-router in order to minimize DB queries seems like a logical enough thing to do.

Then again, it's not like we can minimize our queries to one per e-mail, so we should be careful not to mess up too much anyway.

PR #84, opened 2015-10-17
db-query header.php & syntax

Udera:
Update of db-query in header.php

### DIFF
--- a/vexim/config/header.php
+++ b/vexim/config/header.php
@@ -1,11 +1,13 @@
 <?php
   if (isset($_SESSION['domain_id'])) {
-    $domheaderquery = "SELECT enabled FROM domains WHERE domains.domain_id='" . $_SESSION['domain_id'] . "'";
-    $domheaderresult = $dbh->query($domheaderquery);
-    $domheaderrow = $domheaderresult->fetch();
-    $usrheaderquery = "SELECT enabled FROM users WHERE localpart='" . $_SESSION['localpart'] . "' AND domain_id='" . $_SESSION['domain_id'] . "'";
-    $usrheaderresult = $dbh->query($usrheaderquery);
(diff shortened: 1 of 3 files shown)


vexim/config/header.php
```diff
@@ -13,9 +15,9 @@
     print     "-- " . $_SESSION['domain'] . " ";
   }
   if (isset($_SESSION['domain_id'])) {
-    if (($domheaderrow['enabled'] == "0") || ($domheaderrow['enabled'] == "f")) {
+    if (($domheaderrow['enabled'] === "0") || ($domheaderrow['enabled'] === "f")) {
```
rimas-kudelis: I don't think "f" is actually a valid value here. Integer 0, on the other hand, seems very valid. We should check what is actually returned by the query.

Same for below.

Udera: This "f" was inserted here:
https://github.com/vexim/vexim2/blob/1e807f14b2bd1d9d9cc9b918983e03bcb600c2f6/vexim/config/header.php

Do you see why this was put there?

Udera: Was this put for a new feature, or debugging? The `enabled`-field is either `1` or `0`

Udera:
Ok, I removed this condition completely. I haven't seen where the variable could be 'f', we will see in the final testing, if this causes an error.

rimas-kudelis:
Have you tested if your patch works? In particular, I'm interestėd whether `$domheaderrow['enabled']` and `$usrheaderrow['enabled']` are indeed of string type, and not integers.

Udera:
I checked through the list and used the pull requests #82 #84.
And if you put `if (($domheaderrow['enabled'] === 0))` instead, it won't work, so mysql seems to return everything as string.

rimas-kudelis:
Cool!

PR #85, opened 2015-10-25
apply maxmsgsize as global domain limit and some smaller fixes

Udera:
After some testing, I found:

A global maxmsgsize was not correctly applied.
User couldn't change his preferences (only password). Other little fixes.

### DIFF
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -20,11 +20,11 @@
   $sth = $dbh->prepare($query);
   $sth->execute(array(':domain_id'=>$_SESSION['domain_id']));
   $row = $sth->fetch();
-  if ((isset($_POST['on_avscan'])) && ($row['avscan'] = 1)) {$_POST['on_avscan'] = 1;} else {$_POST['on_avscan'] = 0;}
-  if ((isset($_POST['on_spamassassin'])) && ($row['spamassassin'] = 1)) {$_POST['on_spamassassin'] = 1;} else {$_POST['on_spamassassin'] = 0;}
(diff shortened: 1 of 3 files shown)


vexim/userchangesubmit.php
```diff
@@ -32,21 +32,23 @@
...
+    if (validate_password($_POST['clear'], $_POST['vclear'])) {
+      $cryptedpassword = crypt_password($_POST['clear']);
+      $query = "UPDATE users SET crypt=:crypt WHERE user_id=:user_id";
+      $sth = $dbh->prepare($query);
+      $success = $sth->execute(array(':crypt'=>$cryptedpassword, ':user_id'=>$_SESSION['user_id']));
+      if ($success) {
+        $_SESSION['crypt'] = $cryptedpassword;
+        header ("Location: userchange.php?userupdated");
+        die;
+      } else {
```
rimas-kudelis: These three lines seem unnecessary, so should probably be removed.

Udera: Which 3 line out of these 10 lines do you mean?

rimas-kudelis: Lines 45-47

Udera: Yes, wrong password makes no sense. Either we remove it or we put a database-error.

rimas-kudelis: Yeah. I guess we should either use `failupdated=$email_address`, or create a new error message for sql errors.
Or you can just delete these three lines, and let it fail as it does now. :) I still think we have a big rewrite pending anyway.

rimas-kudelis:
Merging this, but I think it could still be improved: correct quota should be calculted from `users.quota` and `domains.quota` on mail delivery, not during mailbox creation. This would allow to leave `users.quota` at zero, which would mean "inherit quota from domain".

PR #92, opened 2015-10-29
Adding empty-password users through a bug in validate_password()

inoveb:
The latest update https://github.com/vexim/vexim2/commit/d06625b80abc6378cb4aa41c8fdbcaec03ce0d82 introduced a bug into the validate_password() function as it returns true for null values. 
E.g.: **siteaddsubmit.php** line **82**
**if (validate_password($_POST['clear'], $_POST['vclear']))** returns *true* if *clear* and *vclear* input fields have been removed from the DOM before submitting.

The patch also includes tests for the function and some small changes to the function and documentation.

### DIFF
--- a/vexim/config/Tests/functionsValidatePasswordTest.php
+++ b/vexim/config/Tests/functionsValidatePasswordTest.php
@@ -45,4 +45,3 @@ class FunctionValidatePasswordTest extends PHPUnit_Framework_TestCase
     }
 }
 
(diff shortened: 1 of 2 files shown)


rimas-kudelis:
So could you squash the last two commits into one? If not, I'll merge as is.

rimas-kudelis:
OK, nevermind.

Issue #88, opened 2015-10-25
Behavior disabled domain/accounts


Udera:
In the testing-procedure, I found this:
https://github.com/vexim/vexim2/blob/master/docs/checklist.txt#L71-L73

Disabled domain -> postmaster can login, user can't (all should be able to login)
Disabled account -> user can login (should fail)

I don't know, where this feature was lost and if there were some security aspects. Do we want this back? @rimas-kudelis

Udera:
remark: This pull request didn't make a difference: https://github.com/vexim/vexim2/pull/84
So we could merge this.

rimas-kudelis:
I suppose we should fix it, especially considering how simple it must be. :)

PR #95, opened 2015-11-03
Deny access to disabled account but accept access to user account whe…

Udera:
…n domain is disabled.

Users should be able to modify their account when the domain is disabled but not when their account is disabled. The header.php automatically checks if the domain is enabled or the useraccount is still valid and generates the message. No need to pass it in line 89.

Fixes https://github.com/vexim/vexim2/issues/88

### DIFF
--- a/vexim/login.php
+++ b/vexim/login.php
@@ -14,22 +14,23 @@
 
 	# construct the correct sql statement based on who the user is
   if ($_POST['localpart'] === 'siteadmin') {
-		$query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
-      WHERE localpart='siteadmin'
-      AND domain='admin'
-      AND username='siteadmin'
-      AND users.domain_id = domains.domain_id";
+    $query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
+    WHERE localpart='siteadmin'
+    AND domain='admin'
+    AND username='siteadmin'
+    AND users.domain_id = domains.domain_id";
   } else if ($AllowUserLogin) {
-		$query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
-      WHERE localpart=:localpart
-      AND users.domain_id = domains.domain_id
-      AND domains.domain=:domain;";
+    $query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled, users.enabled AS userenabled
+    FROM users,domains
+    WHERE localpart=:localpart
+    AND users.domain_id = domains.domain_id
+    AND domains.domain=:domain;";
   } else {
-		$query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
-      WHERE localpart=:localpart
-      AND users.domain_id = domains.domain_id
-      AND domains.domain=:domain
-      AND admin=1;";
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
Sorry, I guess I failed to notice: I don't think users of a disabled domain should be able to log in. Why should they?

Udera:
I don't know the reason, but it is clearly stated here:
https://github.com/vexim/vexim2/blob/master/docs/checklist.txt#L71-L73

Perhaps the idea is to enable end users to access their mailbox even when the domain is disabled (e.g. if you move a domain to a different server, mails couldn't be delivered to the new server if the domain is still active in vexim).

vexim/login.php
```diff
@@ -20,8 +20,8 @@
       AND username='siteadmin'
       AND users.domain_id = domains.domain_id";
   } else if ($AllowUserLogin) {
-		$query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
-      WHERE localpart=:localpart
+		$query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled, users.enabled AS usersenabled
```
rimas-kudelis: No need for `AS`, or at least call the field `userenabled` (without `s`).

vexim/login.php
```diff
@@ -20,8 +20,8 @@
...
+      FROM users,domains WHERE localpart=:localpart
```
rimas-kudelis: Let's add a line-break before `WHERE`.

rimas-kudelis:
Hm, I wonder if updating the checklist instead wouldn't make more sense. But okay, this works as well.

Udera:
If we prefer a different behaviour, we should update the checklist. I was also wondering why you want people to access their settings when the domain is disabled. 

We need to keep this list up-to-date, it's really useful for testing.

rimas-kudelis:
Agreed.

Udera:
So we rather disable the login for users when the domain is disabled? For the admin as well?

What is the exact use case of domain-disabling? A customer didn't pay? Multiple abused accounts? Disable a domain, if there are temporary problems (too much spam?) For domains moving in/out?

rimas-kudelis:
@avleen please tell us. ;)

avleen:
I've spent a week thinking about this, and I'm not sure :-)
There is certainly value, if a domain is disabled, in allowing the users to access their mailboxes.
But, overall, I think this is more likely to present problems. For example, let's say a domain owner wants to disable access to all their mailboxes (there are many legitimate reasons for wanting to do this).. if we allow access to mailboxes in disabled domains, we just broke that expectation.

The alternative is that if we disable a domain and all mailboxes with it, people could lose access to their email. But this is a matter between the mailbox user and the domain owner.

I think the right approach, when disabling a domain, is to disable all access to accounts in that domain too.

Udera:
Thanks for your reply. I will then create a patch that won't allow 
logins to accounts of disabled domains. If someone needs a different 
behavior, it won't be difficult for him to change it back.

As this behavior is different to previous versions, we need to put it in 
the changelogs.

rimas-kudelis:
When fixing indentation, please do that in a separate commit.

vexim/login.php
```diff
@@ -59,39 +60,43 @@
...
-	
+    header ('Location: site.php');
+    die();
+  }
+  if ($row['admin'] === '1') {
+    header ('Location: admin.php');
+    die();
+ }
+ if (($row['userenabled'] === '0')) {
+   header ('Location: index.php');
```
rimas-kudelis: Perhaps redirect it with a param similarly to `?domaindisabled`?

Udera: I think there is a snippet in the index.php which already adds this 
information (account disabled).

rimas-kudelis:
eh?

Udera:
No, it was header.php

vexim/login.php
rimas-kudelis: But that is also true for disabled domains.
Should we even set the session variables in these cases?

Udera:
I put a new pull request: https://github.com/vexim/vexim2/pull/99

Issue #86, opened 2015-10-25
Mark disabled domains/account


Udera:
In the list of domains or accounts, you won't see which one is not activated. You need to open the preferences. It would be better to mark them in a way, make them italic, other color...

PR #96, opened 2015-11-03
Color disabled accounts/domains red

Udera:
Fixes https://github.com/vexim/vexim2/issues/86

### DIFF
--- a/vexim/adminalias.php
+++ b/vexim/adminalias.php
@@ -59,7 +59,7 @@
           print '<td class="check">';
           print '</tr>';
         }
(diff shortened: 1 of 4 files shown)


rimas-kudelis:
I'd go for gray, not red. ;) You don't disable things to make them more prominent, you disable because you no longer need them.

Udera:
I thought more of a temporary state. If you really don't need the domain/account, you can also delete it. Red is perhaps too alerting, since it's not a real error/problem of the system.

rimas-kudelis:
Yeah, I think red is way too prominent.

Udera:
I used a dark grey (color: #666).

PR #99, opened 2015-11-14
 Deny access to disabled account/domain. Fixes #88

### LINKED ISSUE #88: Behavior disabled domain/accounts

Udera:
Replacing https://github.com/vexim/vexim2/pull/95 to fix https://github.com/vexim/vexim2/issues/88

I put the checks before the session-variable is initialized. Indentation fixes are in a separate commit.

### DIFF
--- a/vexim/config/header.php
+++ b/vexim/config/header.php
@@ -7,12 +7,12 @@
     if ($headersuccess) {
       $headerrow = $headerresult->fetch();
       if ($headerrow['domain'] === "0") {
-        unset($_SESSION['domain_id']);
+        invalidate_session();
         header ("Location: index.php?domaindisabled");
 	die();
       }
       if ($headerrow['user'] === "0") {
-        unset($_SESSION['domain_id']);
+        invalidate_session();
         header ("Location: index.php?userdisabled");
         die();
       }
--- a/vexim/config/header.php
+++ b/vexim/config/header.php
@@ -1,26 +1,5 @@
 <?php
-  if (isset($_SESSION['domain_id'])) {
-    $domheaderquery = "SELECT enabled FROM domains WHERE domains.domain_id=:domain_id";
-    $domheaderresult = $dbh->prepare($domheaderquery);
-    $domsuccess = $domheaderresult->execute(array(':domain_id'=>$_SESSION['domain_id']));
-    if ($domsuccess) { $domheaderrow = $domheaderresult->fetch(); }
-    $usrheaderquery = "SELECT enabled FROM users WHERE localpart=:localpart AND domain_id=:domain_id";
-    $usrheaderresult = $dbh->prepare($usrheaderquery);
-    $usrsuccess = $usrheaderresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
-    if ($usrsuccess) { $usrheaderrow = $usrheaderresult->fetch(); }
-  }
-
-  print "<div id=\"Header\"><p><a href=\"https://github.com/avleen/vexim2\" target=\"_blank\">" . _("Virtual Exim") . "</a> ";
-  if (isset($_SESSION['domain'])) {
-    print     "-- " . $_SESSION['domain'] . " ";
-  }
-  if (isset($_SESSION['domain_id'])) {
-    if ($domheaderrow['enabled'] === "0") {
-      print   _("-- domain disabled (please see your administrator).");
-    } else if ($usrheaderrow['enabled'] === "0") {
-      print   _("-- account disabled (please see your administrator).");
-    }
-  }
+  print "<div id=\"Header\"><p><a href=\"https://github.com/vexim/vexim2\" target=\"_blank\">" . _("Virtual Exim") . "</a> ";
   // First a few status messages about account maintenance
   if (isset($_GET['added'])) {
     printf (_("-- %s has been successfully added."), $_GET['added']);
@@ -68,6 +47,8 @@
     printf (_("-- Deletion of %s was canceled."), $_GET['canceldelete']);
   } else if (isset($_GET['domaindisabled'])) {
     print   _("-- This domain is currently disabled. Please see your administrator.");
+  } else if (isset($_GET['userdisabled'])) {
+    print   _("-- This account is currently disabled. Please see your administrator.");
   } else if (isset($_GET['maxaccounts'])) {
     print   _("-- Your Domain Account Limit Has Been Reached. Please contact your administrator.");
   } else if (isset($_GET['quotahigh'])) {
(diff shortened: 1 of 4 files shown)


rimas-kudelis:
Other than the four empty consecutive lines, the patch looks OK to my tired eyes.

However, I wonder what happens if the account or domain becomes deactivated while the user is still logged in. Should we account for that as well?

rimas-kudelis:
Also, it would be better if we didn't replace PR's, but update them instead. To update, you just have to push to the same branch, and if rewriting commit history, you should use `-f` when pushing.

Udera:
I thought about the active sessions. But the point is, once an account/domain is disabled, exim won't handle mails any more and imap-logins are disabled as well. A user could only change some settings that get active once his account his activated again.

I created a new branch because I wanted to keep the previous changes. I could have pushed the changes of my local new branch to the same remote branch on github.

Udera:
I didn't see consecutive empty lines?

I inserted a forced logout for disabled domains/users who are currently logged in.

vexim/config/header.php
```diff
@@ -1,24 +1,25 @@
...
-    $domsuccess = $domheaderresult->execute(array(':domain_id'=>$_SESSION['domain_id']));
-    if ($domsuccess) { $domheaderrow = $domheaderresult->fetch(); }
-    $usrheaderquery = "SELECT enabled FROM users WHERE localpart=:localpart AND domain_id=:domain_id";
-    $usrheaderresult = $dbh->prepare($usrheaderquery);
-    $usrsuccess = $usrheaderresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
-    if ($usrsuccess) { $usrheaderrow = $usrheaderresult->fetch(); }
-  }
-
-  print "<div id=\"Header\"><p><a href=\"https://github.com/avleen/vexim2\" target=\"_blank\">" . _("Virtual Exim") . "</a> ";
+  print "<div id=\"Header\"><p><a href=\"https://github.com/vexim/vexim2\" target=\"_blank\">" . _("Virtual Exim") . "</a> ";
```
rimas-kudelis: Why did you put output above `header()` calls?  I don't see why it would matter for the application logic, so it looks like the lines are reshuffled just for the sake of it. Please don't do this.

vexim/config/header.php
```diff
@@ -1,24 +1,25 @@
...
-      print   _("-- account disabled (please see your administrator).");
+  if (isset($_SESSION['domain_id'])) { // Check if accound or domain haven't been disabled
+    $headerquery = "SELECT domains.enabled AS domain, users.enabled AS user FROM users,domains
+                    WHERE users.localpart=:localpart AND domains.domain_id=:domain_id AND users.domain_id=domains.domain_id;";
+    $headerresult = $dbh->prepare($headerquery);
+    $headersuccess = $headerresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
+    if ($headersuccess) {
+      $headerrow = $headerresult->fetch();
+      if ($headerrow['domain'] === "0") {
+        unset($_SESSION['domain_id']);
```
rimas-kudelis: I suppose it would be better to do here what `logout.php` does, only the `header()` call should remain different. Perhaps we should even have it always done in a single place, e.g. in `functions.php`?

vexim/config/header.php
Udera: Sorry, I wanted to do it differently in the beginning where the order did matter. I'll change it back.

Udera:
I put a new function invalidate_session() into the config/functions.php to close a session. Therefore the logout.php was updated and I put this function into the config/header.php.

vexim/config/header.php
```diff
@@ -1,26 +1,27 @@
...
-    $usrheaderresult = $dbh->prepare($usrheaderquery);
-    $usrsuccess = $usrheaderresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
-    if ($usrsuccess) { $usrheaderrow = $usrheaderresult->fetch(); }
+  if (isset($_SESSION['domain_id'])) { // Check if accound or domain haven't been disabled
+    $headerquery = "SELECT domains.enabled AS domain, users.enabled AS user FROM users,domains
+                    WHERE users.localpart=:localpart AND domains.domain_id=:domain_id AND users.domain_id=domains.domain_id;";
+    $headerresult = $dbh->prepare($headerquery);
+    $headersuccess = $headerresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
+    if ($headersuccess) {
+      $headerrow = $headerresult->fetch();
```
rimas-kudelis: Hm, what happens if `$headersuccess` or `$headerrow` is false for some reason?

vexim/config/header.php
```diff
@@ -1,26 +1,27 @@
...
-  if (isset($_SESSION['domain_id'])) {
-    $domheaderquery = "SELECT enabled FROM domains WHERE domains.domain_id=:domain_id";
-    $domheaderresult = $dbh->prepare($domheaderquery);
-    $domsuccess = $domheaderresult->execute(array(':domain_id'=>$_SESSION['domain_id']));
-    if ($domsuccess) { $domheaderrow = $domheaderresult->fetch(); }
-    $usrheaderquery = "SELECT enabled FROM users WHERE localpart=:localpart AND domain_id=:domain_id";
-    $usrheaderresult = $dbh->prepare($usrheaderquery);
-    $usrsuccess = $usrheaderresult->execute(array(':localpart'=>$_SESSION['localpart'], ':domain_id'=>$_SESSION['domain_id']));
-    if ($usrsuccess) { $usrheaderrow = $usrheaderresult->fetch(); }
+  if (isset($_SESSION['domain_id'])) { // Check if accound or domain haven't been disabled
```
rimas-kudelis: This comment has two typos. :) Maybe we just don't need it?

vexim/config/header.php
Udera: I'm not sure if I understand correctly. Do you mean we should make a query like `SELECT COUNT(*) FROM users,domains WHERE domains.enabled='1' AND users.enabled='1' AND ...` and we continue only if the query can be made and the result is `1`? For anything else we end the session?

vexim/config/header.php
Udera: OMG, `d` and `t` are not even close on the keyboard for same lame excuse ;-)

vexim/config/header.php
rimas-kudelis: Nah, I mean we should probably have an `if (!$headersuccess)` case to handle possible issues, and then also check if `$headerrow` is not `false` (or is an array) before using it as an array.

Udera:
I added a verification for the database queries.

rimas-kudelis:
Thank you!

PR #100, opened 2015-11-15
Ability to switch AV and SA for all users on/off. Fixes #69

### LINKED ISSUE #69: Allow enabling spamassassin and anti-virus for all domain users
We have disabling already, now we should allow enabling it globally as well. See #66.

Kami-no:
Ability for domain admin to force AV and SA for all users to enable or
disable. Fixes #69

### DIFF
--- a/vexim/sitechange.php
+++ b/vexim/sitechange.php
@@ -124,18 +124,28 @@ include_once dirname(__FILE__) . "/config/httpheaders.php";
                 <td><input name="submit" size="25" type="submit" value="<?php echo _("Submit Changes"); ?>"></td>
             </tr>
         </form>
-	<form name="sadisable" method="post" action="sitechangesubmit.php">
-	  <tr><td><?php echo _("Disable SpamAssassin for all domain users") . "<br><b>(" . _("Warning: cannot be reversed!") . ")</b>:</td>\n"; ?>
-	      <td><input name="sadisable" type="hidden" value="sadisable">
-		  <input name="domain_id" type="hidden" value="<?php print $_GET['domain_id']; ?>">
-		  <input name="submit" type="submit" value="<?php echo _("Disable"); ?>"></td></tr>
-	</form>
-	<form name="avdisable" method="post" action="sitechangesubmit.php">
-	  <tr><td><?php echo _("Disable Anti-Virus for all domain users") . "<br><b>(" . _("Warning: cannot be reversed!") . ")</b>:</td>\n"; ?>
-	      <td><input name="avdisable" type="hidden" value="avdisable">
-		  <input name="domain_id" type="hidden" value="<?php print $_GET['domain_id']; ?>">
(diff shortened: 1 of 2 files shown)


rimas-kudelis:
@Kami-no would you please make a PR without whitespace changes, or at least split it into two commits? Thanks.

Because right now the true essence of this PR can hardly be seen among numerous whitespace and line-break adjustments.

Kami-no:
I'll separate sitechange.php but sitechangesubmit.php modified almost completely for 'case' logic.

vexim/sitechange.php
```diff
@@ -1,86 +1,153 @@
...
+                        $query = "SELECT localpart,domain FROM users,domains
+                            WHERE domains.domain_id=:domain_id
+                            AND admin=1 AND users.domain_id=domains.domain_id";
+                        $sth = $dbh->prepare($query);
+                        $sth->execute(array(':domain_id' => $_GET['domain_id']));
+                        while ($row = $sth->fetch()) {
+                            print '<option value="' . $row['localpart'] . '">' . $row['localpart'] . '</option>' . "\n\t";
+                        }
+                        ?>
+                    </select>
```
rimas-kudelis: The `@` character is still missing, I suppose it's not on purpose?

rimas-kudelis: Actually, you probably just shouldn't change this particular line.

Udera:
I can't disable anti-virus for all users without touching spamassassin.

Settings for the `maximum message size` are ignored (they are updated, but the user can still choose a larger value). I am not sure, if this issue existed before.

After a rewrite of the whole script, it's probably a good idea to verify if all is working as it should: https://github.com/vexim/vexim2/blob/master/docs/checklist.txt

Udera:
Ok, I found the problem with the `maxmsgsize`, it was independent from this pull request. So don't bother for it (@Kami-no), I created a separate pull request: https://github.com/vexim/vexim2/pull/101

It remains the issue with disabling antivirus for all users (without touching spamassassin). Problem should be here: https://github.com/vexim/vexim2/pull/100/files#diff-6449e75348218b3427547d3aafb7a947R28

Udera:
Looks good but I didn't test again.

Udera:
I did test it. It seems to do what it should do. 

* I would change the position of the lower submit button: it should be on the right hand side. Can you make a little spacing that it gets clearer that you need to push the second submit button? (perhaps like on the top, insert a new subheader "Modify SA/AV setting for all accounts".

* If you submit the lower form, there is no field `domain`, so the variable `$_POST['domain']` in https://github.com/Kami-no/vexim2/blob/master/vexim/sitechangesubmit.php#L124 is unknown. Either add a hidden field in the form or get the domain from the db-query.

* If spamassassin is enabled and we disable it, shouldn't we disable it automatically for all accounts?

Udera:
I fixed some stuff and created a new pull request: https://github.com/vexim/vexim2/pull/123

PR #101, opened 2015-12-25
Maximum messages size parameter was ignored for adminuserchange.

Udera:
Repaired the max-massage-size problems discovered during tests of https://github.com/vexim/vexim2/pull/100#issuecomment-166034354.

When an adminuser changed the message size for a user, the `maxmsgsize` was ignored. I added a control.

A further issue was discovered for users changing their settings (userchangesubmit), the `maxmsgsize` was working for a domain with limited massage size (`maxmsgsize>0`) but when the limit was 0 (unlimited), the user could set an individual limit.

### DIFF
--- a/vexim/adminuserchangesubmit.php
+++ b/vexim/adminuserchangesubmit.php
@@ -15,7 +15,7 @@
   }
  
   # Fix the boolean values
-  $query = "SELECT avscan,spamassassin,pipe,uid,gid,quotas
+  $query = "SELECT avscan,spamassassin,pipe,uid,gid,quotas,maxmsgsize
     FROM domains
     WHERE domain_id=:domain_id";
   $sth = $dbh->prepare($query);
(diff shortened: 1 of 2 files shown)


vexim/adminuserchangesubmit.php
```diff
@@ -101,6 +101,12 @@
     $forwardaddr = $_POST['forward'];
   }
 
+  if (isset($_POST['maxmsgsize']) && $row['maxmsgsize']!=='0') {
+    if ($_POST['maxmsgsize']<=0 || $_POST['maxmsgsize']>$row['maxmsgsize']) {
```
rimas-kudelis: I suppose you'd want to add ` || $_POST['maxmsgsize'] == 0` to this too?

Udera: Isn't this case covered by `$_POST['maxmsgsize']<=0`?

rimas-kudelis: Oh snap, indeed. Shame on me!

PR #102, opened 2015-12-25
Remove both spam headers in single 'headers_remove' in virtual_domains router.

sgf-dma:
Two 'headers_remove' options cause error in exim 4.80 from Debian Wheezy.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -87,6 +87,10 @@ hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
 # Comment this line if you want to disable it, instead of + you can use a different separator.
 VEXIM_LOCALPART_SUFFIX = +*
 
+# if you exim version < 4.83 , define macro below for enabling new
+# 'headers_remove' behavior. See https://github.com/vexim/vexim2/pull/102 .
+#OLD_HEADERS_REMOVE = yes
+
 #Debian: hide mysql_servers = localhost::(/var/run/mysqld/mysqld.sock)/vexim/vexim/CHANGE
 #hide mysql_servers = localhost::(/tmp/mysql.sock)/vexim/vexim/CHANGE
 #hide pgsql_servers = localhost/vexim/vexim/CHANGE
@@ -662,18 +666,28 @@ virtual_domains:
                     {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
                     {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
                     } {X-Spam-Flag: YES\n}{} }
-  headers_remove = ${if or { { <{$spam_score_int}{1} } \
-                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
-                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
-                                  } \
-                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
-                          }  {X-Spam-Score}}
-  headers_remove = ${if or { { <{$spam_score_int}{1} } \
-                            { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
-                                   {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
-                                  } \
-                            } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
-                          }  {X-Spam-Report}}
+  # Whether to use old or new headers_remove behavior.
+  .ifndef OLD_HEADERS_REMOVE
+    headers_remove = ${if or { { <{$spam_score_int}{1} } \
+                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                    } \
+                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                            }  {X-Spam-Score}}
+    headers_remove = ${if or { { <{$spam_score_int}{1} } \
+                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                    } \
+                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                            }  {X-Spam-Report}}
+  .else
+    headers_remove = ${if or { { <{$spam_score_int}{1} } \
+                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
+                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
+                                    } \
+                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                            }  {X-Spam-Score:X-Spam-Report}}
+  .endif
   # using local_part_suffixes enables possibility to use user-"something" localparts
   # which could cause you trouble if you're creating email-adresses with dashes in between.
   .ifdef VEXIM_LOCALPART_SUFFIX
--- a/docs/debian-conf.d/main/00_vexim_listmacrosdefs
+++ b/docs/debian-conf.d/main/00_vexim_listmacrosdefs
@@ -2,6 +2,10 @@
 ### main/00_vexim_listmacrosdefs
 #################################
 
+# if you exim version < 4.83 , define macro below for enabling new
+# 'headers_remove' behavior. See https://github.com/vexim/vexim2/pull/102 .
+#OLD_HEADERS_REMOVE = yes
+
 hide mysql_servers = localhost::(/var/run/mysqld/mysqld.sock)/vexim/vexim/CHANGE
 
 # domains
--- a/docs/debian-conf.d/router/250_vexim_virtual_domains
+++ b/docs/debian-conf.d/router/250_vexim_virtual_domains
@@ -55,18 +55,28 @@ virtual_domains:
                     {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
                     {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
                     } {X-Spam-Flag: YES\n}{} }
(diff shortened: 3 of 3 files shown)


Udera:
It was this before: https://github.com/vexim/vexim2/blob/f63539df92285ae793d1aa3fc5a0f953cb351f38/docs/configure
But it didn't work with more recent versions (exim > 4.82):
https://github.com/vexim/vexim2/issues/81
So it was fixed: https://github.com/vexim/vexim2/pull/82

and I thought it to be backward-compatible. Would be nice to get it working in both versions.

Udera:
More than one `headers_remove` seems to be supported since exim 4.83. There are quite some fixes for the `headers_remove` in different 4.8x versions: https://github.com/Exim/exim/blob/master/doc/doc-txt/ChangeLog

For debianic systems, there is already quite a range of different versions (still supported and the next stable version)

| Distribution | Exim version |
| -------------- | ---------------- |
| Debian 7 | 4.80 |
| Debian 8 | 4.84 |
| Debian 9 | 4.86 |
| Ubuntu 12.04 | 4.76 |
| Ubuntu 14.04 | 4.82 |
| Ubuntu 16.04 | 4.86 |

I did test the latest version of vexim with Debian 8. In Debian 7, it's the old version. So we need to verify with Ubuntu 14.04. We use the new version as default and comment the old version (@RimasKudelis)?

ikonia:
I think you've got to target the long term release for this sort of thing eg: Ubuntu 12.04/14.04/16.04 rhel/centos 5,6,7 as these are the ones that people will run mail servers on, so vexim needs to be compatible with these while they are in support in my view

rimas-kudelis:
> So we need to verify with Ubuntu 14.04. We use the new version as default and comment the old version (@rimas-kudelis)?

Yep, I guess so.

Udera:
@sgf-dma Would you mind to change your pull request?
Instead of uncommenting, it is perhaps possible to use a switch (like the `.ifdef`) considering the exim version, there is a variable `$exim_version` (http://www.exim.org/exim-html-current/doc/html/spec_html/ch-variable_index.html).

sgf-dma:
Can you clarify, how can i check version with .ifdef ? As i understand,
.ifdef just checks whether macro is defined or not, but it does not allow to
check macro value.

Variable $exim_version is a version of build, not a version of exim, and,
more importantly, it was introduced between 4.85 and 4.86:

    $ git blame -L 530,530 ./src/src/expand.c
    71224040 (Jeremy Harris 2014-11-29 19:52:09 +0000 530)   { "exim_version",        vtype_stringptr,   &version_string },
    
    $ git log --oneline -1 71224040..exim-4_85
    $ git log --oneline -1 71224040..exim-4_86
    fef95db Doc: parallel builds (make -j) work

Though, in older exim versions there is $version_number variable, which may be used, i
think.

But the question "how" still remains.

Udera:
No, I have no idea how to do this. You could define a variable to choose the old or new router. Or you just uncomment one version. It needs to be documented any way.

sgf-dma:
I've rebased branch and added ifdef, as you suggest. Please, test, because i don't have working vexim configuration now.

Udera:
Thanks for adding this @sgf-dma . It would be great if you could add this also in the separated `vexim2/docs/debian-conf.d`-files.
Regarding the remaining pull-requests for `vexim2.3`, we should perhaps test them together. What do you think @rimas-kudelis ?

docs/configure
```diff
@@ -87,6 +87,20 @@ hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
...
+# Debian 7	      4.80    old
+# Debian 8	      4.84    new
+# Debian 9	      4.86    new (not tested)
+# Ubuntu 12.04        4.76    old (not tested)
+# Ubuntu 14.04        4.82    old (not tested)
+# Ubuntu 16.04        4.86    new (not tested)
+# RHEL 6*             4.72    old (not tested)
+# RHEL 7*             4.84    new (not tested)
+# OpenSUSE 13.x       4.85    new (not tested)
+# OpenSUSE tumbleweed 4.85    new (not tested)
```
rimas-kudelis: Why do we need this list? Wouldn't it be enough to specify which Exim versions which setup applies to?

docs/configure
```diff
@@ -87,6 +87,20 @@ hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
...
+#NEW_HEADERS_REMOVE = yes
```
rimas-kudelis: I think I'd name the macro `OLD_HEADERS_REMOVE` instead. And new syntax should be the default.

rimas-kudelis:
> Regarding the remaining pull-requests for `vexim2.3`, we should perhaps test them together. What do you think @rimas-kudelis?

I'm not exactly sure what you mean.

Udera:
> Regarding the remaining pull-requests for vexim2.3, we should perhaps test them together. What do you think @rimas-kudelis?
> 
> I'm not exactly sure what you mean.

Just if we don't want to wait until these pull requests are completely finished and merged to master, we could create a testing branch. I didn't want to start testing the current `master` with the 4 pull requests still open. When we can get them merged soon, it's fine for me as well.

sgf-dma:
I've fixed all your suggestions.

And can you, please, tell me, when i just rebase a branch and does not write anything in comments, has notification been sent?

Udera:
On 2016-01-19 12:31, sgf-dma wrote:
> I've fixed all your suggestions.
> 
> And can you, please, tell me, when i just rebase a branch and does not
> write anything in comments, has notification been sent?
> 
> --
> Reply to this email directly or view it on GitHub [1].
> 
> 	*
> 
> Links:
> ------
> [1] https://github.com/vexim/vexim2/pull/102#issuecomment-172823670

I got only a notification for your comment.

Z3po:
I've tested again my idea with having just a "headers_add" statement instead of different "headers_remove" statements. Please try my suggestion at _https://github.com/Z3po/vexim2/tree/issue#102_

I've added a exim.filter file which must be placed within the /etc/exim4 directory.
The system_filter is run only once per message and it has not changed in its usability ("headers remove" can be used with colon seperated headers): _http://www.exim.org/exim-html-current/doc/html/spec_html/ch-systemwide_message_filtering.html_

I've done a quick test on my wheezy system and it does what it should do...can anyone test with jessie/new exim version please?

Z3po:
I've just recognized i messed up my commit. i just rebased and pushed it again (if you're already looking to a "old" version the ne commit is "e64700a".

Z3po:
Any concerns with that approach?
If not I will create a pull request against the current master-branch.

Udera:
I'm not sure if we should mix these topics. I haven't really looked into the exim filters, but this doesn't sound good.
> Warning: Because the system filter runs just once, variables that are specific to individual recipient addresses, such as $local_part and $domain, are not set, and the “personal” condition is not meaningful.

But we need to decide based on a user's preference when the spam headers are expected in his mails.

Z3po:
That's not an issue here. We're *removing* all spamassassin headers _once_ for all message recipients (We don't need headers from other spamassassin runs of different mailsystems) and afterwards we're adding them _by recipient_ in the actual transport (where all the user information is available). It's not a problem using filters here as running once for an incoming message is exactly what we like to achieve.

Please check the mentioned commit.

Udera:
I tested this pull request (except Debian 7 which was probably tested by @sgf-dma ):

| Distribution | Exim | old config* | new config |
| ---- | ---- | -----| ----|
| Debian 8 | 4.84 | no | yes |
| Debian 7 | 4.80 | yes | no |
| Ubuntu 14.04 | 4.82 | yes | yes |

*this line is not commented `OLD_HEADERS_REMOVE = yes`

it's surprising that `4.82` supports both version. It seems like ubuntu patched their exim version in `4.82~rc1-1` (https://launchpad.net/ubuntu/trusty/+source/exim4/+changelog).

@Z3po : you have just a different solution, it doesn't solve the problem. You add several lines at one time via `header_add`: https://github.com/Z3po/vexim2/commit/e64700a#diff-b40c611879710c5e4d8a74a79ccce135R854
I thought `headers_add` has the same issues as `headers_remove`.

rimas-kudelis:
@Udera: no. [Exim Spec, Chapter 15](http://www.exim.org/exim-html-current/doc/html/spec_html/ch-generic_options_for_routers.html) explicitly states that:
> Unlike most options, `headers_add` can be specified multiple times for a router; all listed headers are added.

Udera:
But it says the same thing about `headers_remove`
> Unlike most options, headers_remove can be specified multiple times for a router; all listed headers are removed.

rimas-kudelis:
@Udera: hm, you're right. However, `headers_add` [works completely differently](http://www.exim.org/exim-html-4.80/doc/html/spec_html/ch-message_processing.html#SECTheadersaddrem): it doesn't split strings by colons, as `headers_remove` does, but instead allows specifying multiple headers separated by a line-break, and I think THAT was always this way.

At least to me, @Z3po's commit seems to make the code a tiny bit more elegant than our current approach, don't you think? If we're not willing to test and integrate it into v2.3, I think we should definitely look into it for our next version.

Udera:
It's possible that there are less problems, but you need to test it (it takes some time to check this with all different versions, mails with spam-tag, mails without, mailman and so on). And there are small changes to the behavior, e.g. for forwarded mails, we currently remove the `X-Spam-Report` and the `X-Spam-Score` but we keep the `X-Spam-Flag` that would allow users to filter the mail (on an external mailbox). Also this `exim.filter` remains a bit unclear, this isn't used anywhere. If someone is willing to test, I won't object.

I would prefer to get all stuff fixed and release vexim2.3 as soon as possible. All users can then benefit from the changes we did over the last months and all will have the same code base (and not everyone his own little patches and bugfixes). We then can either consider some more changes for a version 2.3.1 or we directly move on to vexim3.

rimas-kudelis:
Alright. So this is ready for merging then?

Udera:
I didn't have problem so far, I don't think we shouldn't wait for anyone to test with Ubuntu 12.04, CentOS or other old OSs. When we merged all the open issues, we can do a final testing and install a new system from scratch to see if it works.

PR #106, opened 2016-02-09
Allow custom usernames.

chusopr:
We typically use vexim with customers that want to upgrade their existing ancient mail system, so we want to keep the same usernames they were using and, most of the times, they use a specific schema for usernames that they want to keep for new users (and that doesn't match e-mail address), so we need to be able to select what username to use when creating a new account.
This patch allows domain admins to select which username to use for new accounts. I don't know if you may be interested in merging it upstream, so I'm pushing it and feel free to reject it if you are not interested.

### DIFF
--- a/vexim/adminaliasadd.php
+++ b/vexim/adminaliasadd.php
@@ -29,6 +29,11 @@
             <td><input name="realname" type="text" class="textfield"></td>
           </tr>
           <tr>
+            <td><?php echo _('User Name'); ?>:</td>
+            <td>
+              <input name="username" type="text" class="textfield"></td>
+          </tr>
+          <tr>
             <td><?php echo _('Address'); ?>:</td>
             <td>
               <input name="localpart" type="text" class="textfield">@
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -45,7 +45,7 @@
 
   # check_user_exists() will die if a user account already exists with the same localpart and domain id
   check_user_exists(
-    $dbh,$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
+    $dbh,$_POST['username'],$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
   );
 
   if ((preg_match("/['@%!\/\|\" ']/",$_POST['localpart']))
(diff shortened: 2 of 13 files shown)


rimas-kudelis:
I have an idea. Why not make custom usernames a config option which could be enabled when necessary? Perhaps this could even be extended by allowing to choose from multiple predefined username schemas, and allowing to use a custom one if none of these predefined schemas work.

This way the "User name" field should only be displayed when custom usernames are selected instead of being always visible and always required. What do you think?

Udera:
On 2016-02-09 11:52, Rimas Kudelis wrote:
> I have an idea. Why not make custom usernames a config option which
> could be enabled when necessary? Perhaps this could even be extended
> by allowing to choose from multiple predefined username schemas, and
> allowing to use a custom one if none of these predefined schemas work.
> 
> 
> This way the "User name" field should only be displayed when custom
> usernames are selected instead of being always visible and always
> required. What do you think?
> 
> --
> Reply to this email directly or view it on GitHub [1].
> 
> 	*
> 
> Links:
> ------
> [1] https://github.com/vexim/vexim2/pull/106#issuecomment-181814392
I like the idea to allow a customizable user names.

With this feature, we should retire the split username field during 
login. The dropdown-menu reveals information about the server (can be 
disabled) and it makes no sense for other username-schemes. Instead, I 
would just use a singe username field. For webmail and IMAP/POP you 
already do it that way.

rimas-kudelis:
Yeah, why not.
@chusopr are you up for extending your patch? :)

chusopr:
Hi, sorry, I'm not very available right now. I'm moving to another country and changing my job, so I am in the middle of a lot of procedures.
Actually, this patch is a change I made months ago for internal use that I decided now to push upstream while I'm cleaning up my projects.

ikonia:
I really don't think custom user names are a priority for a release, I'd rather see less risk / change on optional features, and stabilisation of the task list for road to release before moving forward with more additional features 

Matt

runout-at:
I came around this issue on implementing a new WebUI in Python. (see https://github.com/vexim/vexim2/wiki/WebUI-GUI-in-Flask-Python3)

My idia is to use the username field only for login. In every place where the mail address of a user is needed, it should be constructed from concat_ws('@', users.localpart, domains.domain).
I didn't check if this would work properly with MDA logins (dovecot, courier,...)

PR #107, opened 2016-02-09
Fix groups receipts

chusopr:
If you are using a user table where the username of each user is not their e-mail address (for example #106), groups won't work because provided exim configuration resolves group addresses to the usernames of their members, instead of their e-mail addresses. This works when their usernames are their e-mail addresses, but will fail otherwise.
This patch also fixes forms that assume that username = e-mail address to correctly show e-mail addresses.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -701,7 +701,7 @@ virtual_dom_groups:
                                             d.domain_id = g.domain_id and g.enabled = '1' and \
                                             g.name = '${quote_mysql:$local_part}'}}} \
                  {$sender_address} \
-                 {${lookup mysql{select u.username \
+                 {${lookup mysql{select concat_ws('@', localpart, d.domain) \
                                  from domains d, groups g, group_contents c, users u \
                                  where d.enabled = '1' and d.domain = '${quote_mysql:$domain}' and \
                                        d.domain_id = g.domain_id and g.name = '${quote_mysql:$local_part}' and \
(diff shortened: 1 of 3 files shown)


docs/configure
```diff
@@ -701,7 +701,7 @@ virtual_dom_groups:
                                             d.domain_id = g.domain_id and g.enabled = '1' and \
                                             g.name = '${quote_mysql:$local_part}'}}} \
                  {$sender_address} \
-                 {${lookup mysql{select u.username \
+                 {${lookup mysql{select concat_ws('@', localpart, d.domain) \
```
rimas-kudelis: please use `u.localpart` instead of just `localpart` where SELECTs use data from multiple tables.

vexim/admingroupchange.php
```diff
@@ -106,7 +106,7 @@ class="textfield">
                   </a>
                 </td>
                 <td><?php echo $row['realname']; ?></td>
-                <td><?php echo $row['username']; ?></td>
+                <td><?php echo $row['localpart'].'@'.$_SESSION['domain']; ?></td>
```
rimas-kudelis: I wonder if joining the query with the `domains` table and using the domain name from that table wouldn't be a better idea.

vexim/admingroupchange.php
```diff
@@ -152,7 +152,7 @@ class="textfield">
                 ?>
                   <option value="<?php echo $row['user_id']; 
 					?>"><?php echo $row['realname']; 
-					?>(<?php echo $row['username']; ?>)</option>
+					?> (<?php echo $row['localpart'].'@'.$_SESSION['domain']; ?>)</option>
```
rimas-kudelis: here as well.

docs/configure
chusopr: Done in commit 6e5ef52.

vexim/admingroupchange.php
chusopr: This is what I thought at a first time, but later I saw that it's only allowed to add users in the same domain as the group and that we already have that domain name in `$_SESSION["domain"]` as it's already used in the same file in https://github.com/ultreia-es/vexim2/blob/dafd10b/vexim/admingroupchange.php#L45 so there is no need to add more tables to SQL query.

rimas-kudelis: Okay, that makes sense. Thanks!

PR #110, opened 2016-02-21
Use only username to login like it is used for all other authenticati…

Udera:
…on (IMAP/SMTP)

Follow up for this pull request: https://github.com/vexim/vexim2/pull/106

With the current naming this is just your e-mailaddress (or `siteadmin`) like it is used for all other login processes (IMAP/SMTP).

### DIFF
--- a/vexim/login.php
+++ b/vexim/login.php
@@ -11,6 +11,8 @@
     die;
   }
 
+  if($domainguess === 1 && $_POST['username']!=='siteadmin') $_POST['username'].='@'.preg_replace ("/^mail\./", "", $_SERVER["SERVER_NAME"]);
+
   # sql statement based on username
   $query = "SELECT crypt,username,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled, users.enabled AS userenabled
   FROM users,domains
--- a/vexim/login.php
+++ b/vexim/login.php
@@ -14,7 +14,8 @@
   if($domainguess === 1 && $_POST['username']!=='siteadmin') $_POST['username'].='@'.preg_replace ("/^mail\./", "", $_SERVER["SERVER_NAME"]);
 
   # sql statement based on username
-  $query = "SELECT crypt,username,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled, users.enabled AS userenabled
+  $query = "SELECT users.crypt,users.username,users.user_id,domains.domain,domains.domain_id,users.admin,users.type,
+  domains.enabled AS domainenabled, users.enabled AS userenabled
   FROM users,domains
   WHERE username=:username
   AND users.domain_id = domains.domain_id";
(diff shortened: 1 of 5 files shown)


rimas-kudelis:
I think username guessing should be configurable. While right now it may seem like this configuration step is an overhead, that would change as soon as we add support for custom usernames, which don't have to follow localpart@domainpart schema.
Furthermore, the admin might or might not want this magic in their setup, so it should be possible to not have it enabled, I think.
I'm also unsure if limiting guessed domain name to {current hostname with its leftmost part removed} is satisfactory. I think I'd rather either see some sort of mapping provided via a configuration option, or make it possible to provide the domain name by other means (e.g. via an environment variable). Perhaps a default could be in place, but there should be a way to override it. What do you think?

vexim/login.php
```diff
@@ -5,35 +5,32 @@
...
+      || !isset($_POST['username']) || $_POST['username']===''
   ){
     header ('Location: index.php?login=failed');
     die;
   }
 
   # construct the correct sql statement based on who the user is
-  if ($_POST['localpart'] === 'siteadmin') {
-    $query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
+  if ($_POST['username'] === 'siteadmin') {
```
rimas-kudelis: I think the three cases can and should be merged into one.
Also, we should probably discover whether a user is site admin by checking the `type` field value, not by blindly mapping everyone whose `localpart` matches.

Udera:
There is only one SQL query left, there was no reason to use three different ones. Also added the option back to guess the hostname.

vexim/login.php
```diff
@@ -5,35 +5,21 @@
...
-    $query = "SELECT crypt,localpart,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled FROM users,domains
-    WHERE localpart=:localpart
-    AND users.domain_id = domains.domain_id
-    AND domains.domain=:domain
-    AND admin=1;";
-  }
+  if($domainguess === 1 && $_POST['username']!=='siteadmin') $_POST['username'].='@'.preg_replace ("/^mail\./", "", $_SERVER["SERVER_NAME"]);
+
+  # sql statement based on username
+  $query = "SELECT crypt,username,user_id,domain,domains.domain_id,users.admin,users.type,domains.enabled AS domainenabled, users.enabled AS userenabled
```
rimas-kudelis: Could you please prefix all field names with table names?

vexim/login.php
```diff
@@ -88,13 +73,18 @@
...
 
-  # must be a user, send them to edit their own details
-  header ('Location: userchange.php');
-
+  # must be a user, send them to edit their own details, if User-Login is permitted
+  if($AllowUserLogin===1) {
+    header ('Location: userchange.php');
+    die();
+  } else {
+    header ('Location: index.php?login=failed');
```
rimas-kudelis: Perhaps we should change this to `?login=disabled` or something similar instead?

rimas-kudelis: Also, there no need for the `else` and the curly braces here, I think.

Udera:
Put in all your ideas @rimas-kudelis

rimas-kudelis:
Not all of them. You still left the `else` clause at the very end.

Udera:
Sorry, I thought you wanted to remove `else` with the including `header()` command. I updated the last commit.

rimas-kudelis:
Thanks! I would have wanted a line-break just before that last statement, but didn't want to risk annoying you even more. :)

PR #113, opened 2016-02-21
Smaller fixes on new catchall and userchange

Udera:
Some smaller fixes:

* New catchall, we used `$_POST[user_id]` even though this does not exist
* userchange: when only the `realname` is changed in the upper form, the script runs the other updates and creates errors on non existing `$_POST[value]`-variables.
* userchange: option to `store forwarded mail locally` makes only sense when forwarding is active.

### DIFF
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -3,18 +3,21 @@
   include_once dirname(__FILE__) . "/config/authuser.php";
   include_once dirname(__FILE__) . "/config/functions.php";
   include_once dirname(__FILE__) . "/config/httpheaders.php";
-  if (isset($_POST['on_vacation'])) {$_POST['on_vacation'] = 1;} else {$_POST['on_vacation'] = 0;}
+
+  $_POST['on_vacation'] = isset($_POST['on_vacation']) ? 1 : 0;
   if (isset($_POST['on_forward'])) {
     $_POST['on_forward'] = 1;
     if(!filter_var($_POST['forward'], FILTER_VALIDATE_EMAIL)) {
       header ("Location: userchange.php?invalidforward=".htmlentities($_POST['forward']));
       die;
     }
+    $_POST['unseen'] = isset($_POST['unseen']) ? 1 : 0;
   } else {
     $_POST['on_forward'] = 0;
     $_POST['forward']='';
+    $_POST['unseen']=0;
   }
-  if (isset($_POST['unseen']) && $_POST['on_forward']==1) {$_POST['unseen'] = 1;} else {$_POST['unseen'] = 0;}
+
   # Do some checking, to make sure the user is ALLOWED to make these changes
   $query = "SELECT avscan,spamassassin,maxmsgsize from domains WHERE domain_id=:domain_id";
   $sth = $dbh->prepare($query);
@@ -28,7 +31,7 @@
     }
   }
 
-  if (isset($_POST['realname'])) {
+  if (isset($_POST['realname']) && $_POST['realname']!=="") {
     $query = "UPDATE users SET realname=:realname
 		WHERE user_id=:user_id";
     $sth = $dbh->prepare($query);
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -14,7 +14,7 @@
     $_POST['on_forward'] = 0;
     $_POST['forward']='';
   }
-  if (isset($_POST['unseen'])) {$_POST['unseen'] = 1;} else {$_POST['unseen'] = 0;}
+  if (isset($_POST['unseen']) && $_POST['on_forward']==1) {$_POST['unseen'] = 1;} else {$_POST['unseen'] = 0;}
   # Do some checking, to make sure the user is ALLOWED to make these changes
   $query = "SELECT avscan,spamassassin,maxmsgsize from domains WHERE domain_id=:domain_id";
   $sth = $dbh->prepare($query);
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
I'm not entirely sure about these:
> * userchange: when only the `realname` is changed in the upper form, the script runs the other updates and creates errors on non existing `$_POST[value]`-variables.

Even if you change just the real name, you're still submitting the whole form. If there are unprevented errors due to missing variables, these errors should be fixed instead. 

> * userchange: option to `store forwarded mail locally` makes only sense when forwarding is active.

Then the checkbox in question should be disabled and possibly hidden when forwarding is inactive. But I don't think we should be ignoring changes in this value just because some other value means this one will not be used. Don't you think so?

vexim/userchangesubmit.php
```diff
@@ -28,7 +28,7 @@
     }
   }
 
-  if ($_POST['realname'] !== "") {
+  if (isset($_POST['realname'])) {
```
rimas-kudelis: IMO, checking for non-empty realname was a good thing to do. Perhaps you should use both `isset()` and `!= ''`?

Udera:
> Even if you change just the real name, you're still submitting the whole form. If there are unprevented errors due to missing variables, these errors should be fixed instead.

there are actually 3 different forms.


> Then the checkbox in question should be disabled and possibly hidden when forwarding is inactive. But I don't think we should be ignoring changes in this value just because some other value means this one will not be used. Don't you think so?

I check the vexim-routers, and the `unseen`-option only works if vacation is on. Then we can keep the existing settings. To show this option only when vacation is active, that's probably something for javascript.

rimas-kudelis:
> there are actually 3 different forms.

OK. But in fact, the right way to update the user would be via a single `UPDATE` query, not three different queries run under different circumstances, which require tight coordination between how frontend forms and backend processing are made. But that's probably one of the things to do for Vexim3?

> I check the vexim-routers, and the unseen-option only works if vacation is on. Then we can keep the existing settings. To show this option only when vacation is active, that's probably something for javascript.

Of course we *can* keep the existing settings, but the question is whether we *should*. I think we should respect the changes, if they are being made.
And yes, disabling of the dependent field will have to be done via javascript.

Udera:
>> there are actually 3 different forms.
> 
> OK. But in fact, the right way to update the user would be via a single UPDATE query, not three different queries run under different circumstances, which require tight coordination between how frontend forms and backend processing are made. But that's probably one of the things to do for Vexim3?

We have this problem on nearly every page in current vexim. In some scripts, it would change the logic, so I don't think it's worth to consider such changes in this version. In the new version there is hopefully a more responsive design.



>> I check the vexim-routers, and the unseen-option only works if vacation is on. Then we can keep the existing settings. To show this option only when vacation is active, that's probably something for javascript.
> 
> Of course we can keep the existing settings, but the question is whether we should. I think we should respect the changes, if they are being made.
> And yes, disabling of the dependent field will have to be done via javascript.

The existing setting makes sense: Someone forwards his mail during vacation and keeps a local copy. The next time he activates the forwarder he wants to use the same settings.

The javascript to hide the features unless forwarding is activated could be useful at a number of other pages (e.g. spamassassin points).

rimas-kudelis:
> The existing setting makes sense: Someone forwards his mail during vacation and keeps a local copy. The next time he activates the forwarder he wants to use the same settings.

I still don't understand why we should ignore changes the users make just because these changes cannot have any immediate effect. The only explanation for that that I could come up with is "it makes it more obvious whether forwarding is on or off", but I'm not sure if that is a rationale good enough.

Anyway, looking at `userchangesubmit.php` again, now I see we're doing the same with forward e-mail address, which we also clear when forwarding is off. In that case I suppose we can keep this behavior for now. Would you mind merging your `if` block with the `if` block above?

Udera:
I harmonized the last submit-button and fixed the javascript to focus the first input field as well.

rimas-kudelis:
Yeah, but... Would you mind merging the two `if` blocks in `userchangesubmit.php` so that we don't check the value of `$_POST['on_forward']` twice?

Udera:
Now, I got the `if`-block you were referring to (hopefully) ;-)

rimas-kudelis:
Yep, that's the block I was talking about. But you forgot to paste the `if (isset($_POST['unseen'])) {$_POST['unseen'] = 1;}` piece. :)

Udera:
I was too happy, to find the right place that I forgot about the initial purpose :D

rimas-kudelis:
Am I **that** cryptic with my comments? o_O

Udera:
It's too early, there are too many `if`s, ...

rimas-kudelis:
Would you please put `$_POST['unseen'] = 1;` on its own line, indent it prorperly and add curly braces around it?

Udera:
voilà, with curly brackets :D

rimas-kudelis:
Come to think of it, we should set `unseen` to `0` if it is not set, but forwarding is enabled... I forgot about that before. :/

rimas-kudelis:
I suppose this little `if ` could even be replaced with this:
`$_POST['unseen'] = (int)isset($_POST['unseen'])`

And above that block, there's this piece of code currently:
`if (isset($_POST['on_vacation'])) {$_POST['on_vacation'] = 1;} else {$_POST['on_vacation'] = 0;}`
which could then be replaced with this:
`$_POST['on_vacation'] = (int)isset($_POST['on_vacation'])`

Although in future versions we should probably allow flags to be set to `0` meaning "off".

rimas-kudelis:
Alternative syntax (even better):
`$_POST['unseen'] = isset($_POST['unseen']) ? 1 : 0;`
and
`$_POST['on_vacation'] = isset($_POST['on_vacation']) ? 1 : 0;`

Udera:
If you have more ideas, I'm back online tonight ;-)

rimas-kudelis:
Definitely not in this PR. :)

Thank you once more, I really appreciate your patience!

PR #114, opened 2016-02-21
Always allow siteadmin to login

Udera:
`userenabled` is not queried in https://github.com/vexim/vexim2/blob/master/vexim/login.php#L17 so `$row[userenabled]` doesn't exist. We could add this to the SQL queries but it why should someone disable siteadmin? Via vexim, there is no way to do (or undo) this.

### DIFF
--- a/vexim/login.php
+++ b/vexim/login.php
@@ -63,13 +63,15 @@
     header ('Location: index.php?login=failed');
     die();
   }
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
@Udera: I've merged this in, but you should update #110 to run the check for the `username` field, not `localpart`.

Issue #108, opened 2016-02-14
db, field length too short


runout-at:
The field length if `pop` and `smtp` fields in the `users` table is too short.
It's only 255 bytes which is very less on long user/domain names.
I would suggest 4096 bytes for those fields. AFAIR 4096 is the spec for path length in the file system.

Udera:
`username` doesn't have to be a mail address, so we could allow larger usernames. `pop` and `smtp` don't need to be mail addresses (can be pipe commands), so could be longer as well.

mail addresses have a maximum length of 254 characters:
https://en.wikipedia.org/wiki/Email_address#Syntax
https://tools.ietf.org/html/rfc3696#page-6
we have currently `localpart`=192 (should be 64) and `domain`=64 (should be 255) 
but the total length must not exceed 254 characters (we do not verify that at the moment):
http://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690

runout-at:
yes, but the fields contain the whole path like
/var/opt/vmail/mailserver.my-extra-long-domain-name.at/my-extra-long-username/Maildir
and as you stated maybe a pipe command which can also be longer than 255 char.

i run into that problem by upgrading every time because i already have bigger fields in my database. and the update script shortens the fields.

rimas-kudelis:
Let's patch this!
We probably have to adjust maximum username length for #106 as well.

Udera:
I added an address-check: https://github.com/Udera/vexim2/commit/3ab9e40b6282229b6d98c03daf45cebc0a59028d

@rimas-kudelis: Can you fix the db-fields? This would be part of your db-migration-script.

rimas-kudelis:
Should we shorten `localpart` as well? It doesn't make a lot of sense to have a localpart longer than 60 characters, but it wouldn't be nice to cut it off silently as well. If we do shorten it, I think it should be mentioned in release notes or upgrade notes.

Udera:
We should shorten localpart and put this in the release notes. I thought 
about a small tool to check the existing data base if all users have 
valid addresses. But testing my address validation, I considered that 
you would have to use extremely long mail addresses. That's very 
unlikely an issue for anybody.

Udera:
@rimas-kudelis : Did you update the length of the database-fields? I didn't touch that because you already worked on a few changes in here: https://github.com/vexim/vexim2/pull/72

rimas-kudelis:
[...]

PR #115, opened 2016-02-24
Check length of mail address

Udera:
Check if mail address is RFC conform.
partly fixes https://github.com/vexim/vexim2/issues/108

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -43,6 +43,10 @@
     die;
   }
 
+  check_mail_address(
+    $_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
+  );
+
   # check_user_exists() will die if a user account already exists with the same localpart and domain id
   check_user_exists(
     $dbh,$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
--- a/vexim/config/functions.php
+++ b/vexim/config/functions.php
@@ -43,6 +43,23 @@
         }
     }
 
+    /**
+     * Check if mail address is compliant with RFC 3696.
+     * localpart must not exceed 64 char, and the complete mail address
+     * must not exceed 254 characters.
+     *
+     * @param  string  $localpart
+     * @param  string  $domain
+     * @param  string  $page       page to return to in case of failure
+     */
+    function check_mail_address($localpart,$domain,$page)
+    {
+      if ((strlen($localpart)+strlen($domain)>253) || strlen($localpart)>64)
+        {
+            header ("Location: $page?addresstoolong=$localpart");
+            die;
+        }
+    }
 
     /**
      * Render the alphabet. Directly onto the page.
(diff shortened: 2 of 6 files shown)


vexim/config/functions.php
```diff
@@ -43,6 +43,24 @@ function check_user_exists($dbh,$localpart,$domain_id,$page)
...
+     *
+     * @param  string  $localpart
+     * @param  string  $domain
+     * @param  string  $page       page to return to
+     */
+    function check_mail_address($localpart,$domain,$page)
+    {
+      if ((strlen($localpart)+strlen($domain)>253) || strlen($localpart)>64)
+        {
+            header ("Location: $page?addresstoolong=$localpart");
```
rimas-kudelis: This doesn't seem to account for the possibility of vexim being hosted in a subdirectory.

vexim/config/functions.php
```diff
@@ -43,6 +43,24 @@ function check_user_exists($dbh,$localpart,$domain_id,$page)
         }
     }
 
+    /**
+     * Check if mail address is compliant with RFC 3696.
+     *
+     * Queries database $dbh, and redirects to the $page is the user already
+     * exists.
```
rimas-kudelis: ORLY?

vexim/adminaliasaddsubmit.php
```diff
@@ -43,6 +43,11 @@
     die;
   }
 
+  # check_mail_address() will die if the localpart (>64 char) or the total address (>254 char) is too long
```
rimas-kudelis: This comment seems unnecessary, especially considering that you haven't added it in other places.

vexim/config/functions.php
```diff
@@ -43,6 +43,24 @@ function check_user_exists($dbh,$localpart,$domain_id,$page)
         }
     }
 
+    /**
+     * Check if mail address is compliant with RFC 3696.
```
rimas-kudelis: It's probably worth adding specifying the limits (254 and 64) here.

vexim/config/functions.php
```diff
@@ -43,6 +43,24 @@ function check_user_exists($dbh,$localpart,$domain_id,$page)
...
+     *
+     * Queries database $dbh, and redirects to the $page is the user already
+     * exists.
+     *
+     * @param  string  $localpart
+     * @param  string  $domain
+     * @param  string  $page       page to return to
```
rimas-kudelis: This should say "page to return to in case of failure".

vexim/config/functions.php
Udera: This is the same trick we use in `check_user_exists()`. All accessible pages within vexim reside in the same directory, a simple relative link should be fine. Or do I misunderstand something?

Udera:
Another try :D

vexim/config/functions.php
rimas-kudelis: Uhm, I was somehow under impression that this redirect URL would be treated as absolute, not relative. If this is not the case, I retract my comment.

PR #117, opened 2016-02-25
Fix undef variable, improve domain-alias

Udera:
* At some places, I fixed error caused by `$_POST["localpart"]`
* The destination for alias-domains can be selected from a list (instead of typing the destination domain manually). This way, we don't need to check the destination domain for existence.

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -52,6 +52,10 @@
     $dbh,$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
   );
 
+  if(!isset($_POST['realname']) || $_POST['realname']==='') {
+    $_POST['realname']=$_POST['localpart'];
+  }
+
   if ((preg_match("/['@%!\/\|\" ']/",$_POST['localpart']))
     || preg_match("/^\s*$/",$_POST['realname'])) {
     header("Location: adminalias.php?badname={$_POST['localpart']}");
--- a/vexim/adminfail.php
+++ b/vexim/adminfail.php
@@ -35,6 +35,8 @@
               . '<td align="center">'
               . '<a href="adminfaildelete.php?user_id='
               . $row['user_id']
+              . '&localpart='
+              . $row['localpart']
               . '"><img class="trash" src="images/trashcan.gif" title="'
               . _('Delete fail ')
               . $row['localpart']
--- a/vexim/adminfail.php
+++ b/vexim/adminfail.php
@@ -35,7 +35,7 @@
               . '<td align="center">'
               . '<a href="adminfaildelete.php?user_id='
               . $row['user_id']
-              . '&localpart='
+              . '&amp;localpart='
               . $row['localpart']
               . '"><img class="trash" src="images/trashcan.gif" title="'
               . _('Delete fail ')
--- a/vexim/siteaddsubmit.php
+++ b/vexim/siteaddsubmit.php
@@ -140,15 +140,15 @@
     }
   } else if ($_POST['type'] == "alias") {
     $query = "INSERT INTO domainalias (domain_id, alias)
-              VALUES (:domain_id, :alias)";
+              SELECT domains.domain_id, :alias FROM domains WHERE domains.domain_id=:domain_id";
     $sth = $dbh->prepare($query);
-    $success = $sth->execute(array(':domain_id'=>$_POST['aliasdest'], ':alias'=>$_POST['domain']));
-    if (!$success) {
+    $sth->execute(array(':domain_id'=>$_POST['aliasdest'], ':alias'=>$_POST['domain']));
+    if ($sth->rowCount()!==1) {
       header ("Location: site.php?failaddeddomerr={$_POST['domain']}");
       die;
     } else {
       header ("Location: site.php?added={$_POST['domain']}" .
-              "&type={$_POST['type']}");
+              "&amp;type={$_POST['type']}");
       die;
     }
   } else {
--- a/vexim/siteaddsubmit.php
+++ b/vexim/siteaddsubmit.php
@@ -132,10 +132,6 @@
       } else {
         header ("Location: site.php?added={$_POST['domain']}" .
                 "&type={$_POST['type']}");
-        mail("{$_POST['localpart']}@{$_POST['domain']}",
-              vexim_encode_header(_("Welcome Domain Admin!")),
-              "$welcome_newdomain",
-              "From: {$_POST['localpart']}@{$_POST['domain']}\r\nMIME-Version: 1.0\r\nContent-Type: text/plain; charset=utf-8\r\nContent-Transfer-Encoding: 8bit\r\n");
         die;
       }
     } else {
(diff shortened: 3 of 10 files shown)


Udera:
Found some more problems:
* again `localpart`
* some html errors (inverted tags, misplaced hidden fields, ...)
* creation of alias-address fails if there is no `realname`. However, a `realname` is not really needed only for the admins convenience. Instead of failing, an empty `realname` is set to `localpart`

vexim/adminaliasaddsubmit.php
```diff
@@ -52,6 +52,10 @@
     $dbh,$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
   );
 
+  if($_POST['realname']==='') {
```
rimas-kudelis: I would probably also check for `!isset()`, even if knowing that this field is always being submitted.

vexim/adminuseraddsubmit.php
```diff
@@ -158,7 +158,7 @@
       mail("{$_POST['localpart']}@{$_SESSION['domain']}",
         vexim_encode_header(sprintf(_("Welcome %s!"), $_POST['realname'])),
         "$welcome_message",
-        "From: {$_SESSION['localpart']}@{$_SESSION['domain']}\r\nMIME-Version: 1.0\r\nContent-Type: text/plain; charset=utf-8\r\nContent-Transfer-Encoding: 8bit\r\n");
```
rimas-kudelis: Why? This is an e-mail sent from admin's address. It should stay that way IMO.

vexim/adminfail.php
```diff
@@ -35,6 +35,8 @@
               . '<td align="center">'
               . '<a href="adminfaildelete.php?user_id='
               . $row['user_id']
+              . '&localpart='
```
rimas-kudelis: This should really say `&amp;localpart`, although I'm not sure why we need localpart at all, just to delete an alias by its id.

vexim/config/httpheaders.php
```diff
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
```
rimas-kudelis: Adding a doctype before headers are sent will cause problems with output buffering disabled.
If you really want it, I think you should add it before the `<html>` tag in all pages, although I wonder if it makes any difference with the sort of HTML we're talking about.
And I don't think you can add it after the `header()` lines here as well, because this file is included in `FOOsubmit.php` files as well, which might send their own headers.

vexim/adminuseraddsubmit.php
rimas-kudelis: I mean, if `$_SESSION[`localpart`] is undefined here, we should either define it somewhere (upon successful login?), or use different means to compose admin's email address (e.g. by storing her address in a session variable).

vexim/siteaddsubmit.php
```diff
@@ -132,10 +132,6 @@
       } else {
         header ("Location: site.php?added={$_POST['domain']}" .
                 "&type={$_POST['type']}");
-        mail("{$_POST['localpart']}@{$_POST['domain']}",
```
rimas-kudelis: Why would you drop this?

vexim/siteaddsubmit.php
```diff
@@ -143,25 +139,10 @@
       die;
     }
   } else if ($_POST['type'] == "alias") {
-    $query = "SELECT domain_id FROM domains
```
rimas-kudelis: Shouldn't we test if the domain id exists before using it?

vexim/siteaddsubmit.php
Udera: Because this is a relay domain. For relay domains (neither an alias -> see condition line 83, nor local see condition line 104) we don't know a valid mail address (the principal error was that `$POST[localpart]` is not defined. We could only send a mail to `postmaster@relaydomain` supposing that this address always exists.

vexim/adminfail.php
Udera: I think we used it for the confirmation/error-message.

vexim/adminuseraddsubmit.php
Udera: Sorry, I thought it was the destination. We should put the localpart in `$_SESSION[]`

vexim/config/httpheaders.php
Udera: I ran the html code through a code checker to identify problems (of some pages, there are probably more issues). So we would need to add it to each site manually before the `<html>`-tag? I didn't see any impact on rendering with our fancy code ;-)

vexim/siteaddsubmit.php
Udera: Before the changes, you had to type in the domain name of the destination domain.

Now, you can choose the destination domain from a list (https://github.com/Udera/vexim2/blob/undef/vexim/siteadd.php#L175-L184) and the `domain_id` is submitted.

vexim/config/httpheaders.php
rimas-kudelis: Maybe you have output buffering enabled. I don't remember if that is the default, but I'm quite sure it can be turned off.
The alternative is probably to use `ob_start` and `ob_flush`, but you'd still have to edit almost all of our php scripts to either include a "footer" script of some sort, or to call `ob_flush` explicitly. At this point, just adding a doctype where necessary might be easier (or leaving it out since everything works anyway and we'll be killing all of this markup in our next release anyway).

vexim/adminfail.php
rimas-kudelis: The right way would be to query database and get it from there just before asking.
Anyway, I just want you to escape the `&` character if you'll be pushing any other changes to this branch. If not, this can stay as is.

vexim/adminuseraddsubmit.php
rimas-kudelis: > Sorry, I thought it was the destination. We should put the localpart in `$_SESSION[]`

Or a full email address, since I doubt we're using localpart for anything else than composing the email address. It's up to you.

vexim/siteaddsubmit.php
rimas-kudelis: Oh, if this is indeed a relay domain, then you're right, there's no need to send the email.

vexim/siteaddsubmit.php
rimas-kudelis: Yes, you said it in your PR. But what I'm saying is it's still possible to submit arbitrary numbers, so we should still run a quick check to see if the id is valid instead of just assuming that it is.

Udera: I changed the `INSERT`-query that it gets the `domain_id` as a result from the `domains`-table

vexim/adminaliasaddsubmit.php
```diff
@@ -52,6 +52,10 @@
     $dbh,$_POST['localpart'],$_SESSION['domain_id'],'adminalias.php'
   );
 
+  if(!isset($_POST['realname']) && $_POST['realname']==='') {
```
rimas-kudelis: This will never match, you should use `||`, not `&&`.

vexim/siteaddsubmit.php
rimas-kudelis: Have you tested how it fails? I haven't, but I suspect not gracefully at all. I'd rather see a separate check and redirect to `site.php?baddestdom=$domain_id` and an error message easy to comprehend instead.

Udera: It just didn't do anything. I added a return message in case of error.

PR #118, opened 2016-02-28
Fix typo for css style Menu

Udera:
Just a small typo in a few files.


rimas-kudelis:
I quickly checked, and going lowercase seems to require just three changes as opposed to 20, but well, nevermind. :)

Udera:
For other styles, we used with capital letters, so I preferred to keep them. There are some tools to make search and replace quite fast, so I didn't have to open each file manually :D

rimas-kudelis:
Oh really? Tell me more! :D

PR #122, opened 2016-03-02
Continue #100

Udera:
I continued the work of https://github.com/vexim/vexim2/pull/100

### DIFF
--- a/vexim/sitechangesubmit.php
+++ b/vexim/sitechangesubmit.php
@@ -13,6 +13,9 @@ if (isset($_POST['clear'])) {$action = 'adminpass';}
 // Main action switch
 switch ($action) {
   case 'allusers':
+    // Default value
+    $success = TRUE;
+
     // Process SpamAssassin
     if (isset($_POST['on_spamassassin'])) {
       if ($_POST['on_spamassassin'] != '2' && ($_POST['on_spamassassin'] == '1' || $_POST['on_spamassassin'] == '0')) {
--- a/vexim/sitechangesubmit.php
+++ b/vexim/sitechangesubmit.php
@@ -1,99 +1,126 @@
 <?php
-  include_once dirname(__FILE__) . "/config/variables.php";
-  include_once dirname(__FILE__) . "/config/authsite.php";
-  include_once dirname(__FILE__) . "/config/functions.php";
-  include_once dirname(__FILE__) . "/config/httpheaders.php";
+include_once dirname(__FILE__) . "/config/variables.php";
+include_once dirname(__FILE__) . "/config/authsite.php";
+include_once dirname(__FILE__) . "/config/functions.php";
+include_once dirname(__FILE__) . "/config/httpheaders.php";
 
-  $success = false;
-  if (isset($_POST['sadisable'])) {$action = 'sadisable';}
-  if (isset($_POST['avdisable'])) {$action = 'avdisable';}
-  if (isset($_POST['avscan'])) {$_POST['avscan'] = 1;} else {$_POST['avscan'] = 0;}
-  if (isset($_POST['spamassassin'])) {$_POST['spamassassin'] = 1;} else {$_POST['spamassassin'] = 0;}
-  if (isset($_POST['enabled'])) {$_POST['enabled'] = 1;} else {$_POST['enabled'] = 0;}
-  if (isset($_POST['pipe'])) {$_POST['pipe'] = 1;} else {$_POST['pipe'] = 0;}
-  if ($_POST['max_accounts'] == '') {$_POST['max_accounts'] = '0';}
-  if (isset($_POST['clear'])) {
+$success = false;
+
+// Detect the user action
+if (isset($_POST['allusers'])) {$action = 'allusers';}
+if (isset($_POST['clear'])) {$action = 'adminpass';}
+
+// Main action switch
+switch ($action) {
+  case 'allusers':
+    // Process SpamAssassin
+    if (isset($_POST['on_spamassassin'])) {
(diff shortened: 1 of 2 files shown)


Udera:
I created a new rebased pull request: https://github.com/vexim/vexim2/pull/123

rimas-kudelis:
Please use normal issue titles in future. :P

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
+
+// Main action switch
+switch ($action) {
+  case 'allusers':
+    // Default value
+    $success = TRUE;
+
+    // Process SpamAssassin
+    if (isset($_POST['on_spamassassin'])) {
+      if ($_POST['on_spamassassin'] != '2' && ($_POST['on_spamassassin'] == '1' || $_POST['on_spamassassin'] == '0')) {
```
rimas-kudelis: The two `if`s can be reduced to one here. Also, please use equivalency.

rimas-kudelis: And there's no point in ensuring that the value is not 2 when the next thing you do is ensure it's either 1 or 0.

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
+        $sth = $dbh->prepare($query);
+        $success = $sth->execute(array(
+            ':domain_id' => $_POST['allusers'],
+            ':on_spamassassin' => $_POST['on_spamassassin']));
+      }
+    }
+
+    // Process Anti-Virus
+    if (isset($_POST['on_avscan']) && $success) {
+      if ($_POST['on_avscan'] != '2' && ($_POST['on_avscan'] == '1' || $_POST['on_avscan'] == '0')) {
```
rimas-kudelis: same here

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
+    if (isset($_POST['spamassassin'])) {
+      $_POST['spamassassin'] = 1;
+    } else {
+      $_POST['spamassassin'] = 0;
+      $query = "UPDATE users SET on_spamassassin=:on_spamassassin WHERE domain_id=:domain_id";
+      $sth = $dbh->prepare($query);
+      $sth->execute(array(
+        ':domain_id' => $_POST['domain_id'],
+        ':on_spamassassin' => '0'));
+    }
```
rimas-kudelis: What's the point of having separate `allusers` action if you do the same here? From my understanding, these two blocks are here by mistake, aren't they?

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
 
-  if ($groupinfo = @posix_getgrgid ($gid)) {
-    $gid = $groupinfo['gid'];
-  } elseif ($groupinfo = @posix_getgrnam ($gid)) {
-    $gid = $groupinfo['gid'];
-  } else {
-    header ("Location: site.php?failuidguid={$_POST['domain']}");
-    die;
-  }
+    if ($_POST['max_accounts'] == '') {$_POST['max_accounts'] = '0';}
```
rimas-kudelis: Please add an `isset` check and put the assignment on its own line, just like above.

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
-        ':sa_refuse'=>((isset($_POST['sa_refuse'])) ? $_POST['sa_refuse'] : 0),
-        ':spamassassin'=>$_POST['spamassassin'], ':enabled'=>$_POST['enabled'],
-        ':domain_id'=>$_POST['domain_id'],
-        ));
-  }
+    // User can specify either UID, or username, the former being preferred.
+    // Using posix_getpwuid/posix_getgrgid even when we have an UID is so we
+    // are sure the UID exists.
+    if (isset ($_POST['uid'])) {$uid = $_POST['uid'];}
+    if (isset ($_POST['gid'])) {$gid = $_POST['gid'];}
```
rimas-kudelis: Please put assignments on their own lines, just like above.

vexim/sitechangesubmit.php
```diff
@@ -1,99 +1,145 @@
...
+}
 
-# Just-in-case catchall
-header ("Location: site.php?failupdated={$_POST['domain']}");
-?>
-<!-- Layout and CSS tricks obtained from http://www.bluerobot.com/web/layouts/ -->
+if ($success) {
+  header("Location: site.php?updated={$_POST['domain']}");
+  die;
+} else {
```
rimas-kudelis: No need for the `else` condition here, since we won't reach it anyway if the `if` is satisfied. Just leave the comment and the `header()` call. Although, while here, why not change the comment style to a more common `// `.

vexim/sitechangesubmit.php
Udera: The problem I had was: If you disable spamassassin or avscan for the domain and you don't disable it for all users, exim continues filtering spam/virus-mails. The spam router (https://github.com/vexim/vexim2/blob/master/docs/configure#L518-L525) doesn't verify if spamassassin is enabled for the domain itself.
So in this case, the additional feature to turn on/of SA/AV for all users only makes sense if you enable SA/AV later and want to activate it for all existing users.

Alternative: in the exim-router, we check `domains.on_spamassassin=1`. I would prefer this option, because if you turn of SA for a domain by accident, you can easily turn it on again without the users loosing their settings.

rimas-kudelis: Exactly. We should do the check in the router.

PR #124, opened 2016-03-03
Updated some field lengths. Fixes #108.

### LINKED ISSUE #108: db, field length too short

rimas-kudelis:
New limits are:
- Filesystem paths: 4096 characters (current default limit of the Linux kernel)
- Domain names: 255 characters. Practical limit is in fact 252, but nevermind
- Localparts: 64 characters (per RFC3696, Errata #1690)

### DIFF
--- a/setup/create_db.pl
+++ b/setup/create_db.pl
@@ -183,8 +183,8 @@ sub create_mysqldb {
 
 sub create_mysqltables {
    $mydbh->do("CREATE TABLE IF NOT EXISTS $databasename.domains (domain_id mediumint(8) unsigned NOT NULL auto_increment,
-	domain varchar(64) NOT NULL default '',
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
@Udera please review and merge if happy. ;)

Udera:
The `blocklists` table, shouldn't it be possible to add valid mail address? It is only 192 characters long.

rimas-kudelis:
You mean the `blockval` column, right?

rimas-kudelis:
@Udera: I've updated the PR.

Udera:
Exactly. I will try to merge it ..

PR #125, opened 2016-03-03
Respect domain-wide settings for spamassassin and avscan

Udera:
Only respect the sa/av-settings of a user if it was enabled for the domain.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -508,6 +508,7 @@ ditch_malware:
 		                where localpart = '${quote_mysql:$local_part}' \
 		                and domain = '${quote_mysql:$domain}' \ 
 		                and users.on_avscan = '1' \
+                                and domains.avscan = '1' \
 		                and users.domain_id=domains.domain_id}}}{1} }} {yes}{no} }
 
 # spam drop router
@@ -519,6 +520,7 @@ ditch_spam_drop:
         where localpart = '${quote_mysql:$local_part}' \
         and domain = '${quote_mysql:$domain}' \
         and users.on_spamassassin = '1' \
+        and domains.spamassassin = '1' \
         and users.spam_drop = '1' \
         and users.domain_id=domains.domain_id \
         and users.sa_refuse > 0 }{$value}fail}} {yes}{no}}
@@ -540,6 +542,7 @@ ditch_spam:
         where localpart = '${quote_mysql:$local_part}' \ 
         and domain = '${quote_mysql:$domain}' \ 
         and users.on_spamassassin = '1' \ 
+        and domains.spamassassin = '1' \
         and users.spam_drop = '0' \
         and users.on_forward = '0' \
         and users.type = 'local' \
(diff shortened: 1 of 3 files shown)


docs/configure
```diff
@@ -654,7 +657,8 @@ virtual_domains:
   domains = +local_domains
   address_data = ${lookup mysql{\
         select smtp, users.sa_tag*10 AS sa_tag, users.on_spamassassin AS on_spamassassin, \
-        users.uid AS uid, users.gid AS gid, quota from users,domains \
+        users.uid AS uid, users.gid AS gid, quota, domains.spamassassin AS dom_spamassassin \
+        from users,domains \
```
rimas-kudelis: I think you could select `users.on_spamassassin OR domains.spamassassin as on_spamassassin`. At least MySQL [seems to support](https://dev.mysql.com/doc/refman/5.5/en/logical-operators.html#operator_or) it, but I'm not 100% sure about [PgSQL](http://www.postgresql.org/docs/current/static/functions-logical.html) (their manual only speaks about boolean variables, although I'm convinced `1` and `0` would qualify as well).

docs/configure
```diff
@@ -672,20 +676,20 @@ virtual_domains:
                               { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
                                      {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
                                     } \
-                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
+                              } { eq {0}{${extract{dom_spamassassin}{$address_data}}}} \
```
rimas-kudelis: These three don't seem right. I admit I haven't tried to parse this condition, but I doubt removing `on_spamassassin` is the right thing to do here.

Udera: just translated to php, we would need this:
`if ((spamscore < limit && on_spamassassin==1) || on_spamassassin==0 || dom_spamassassin==0)`
or we could use your nice expression from above. I will have a look into it tomorrow. This is definitely one thing I should test as well.

docs/configure
Udera: nice idea. Shouldn't we use `users.on_spamassassin AND domains.spamassassin as on_spamassassin`?

rimas-kudelis: Correct.

docs/configure
rimas-kudelis: The thing is, you removed `|| on_spamassassin==0` from the condition.
But I think the shorter these conditions, the easier they are to read, so I'm all for the "nice expression" path. :)

rimas-kudelis:
By the way, don't you think we should actually remove these per-domain controls? I see no reason at all why anyone would want to disable spam and antivirus checks for an entire domain they are hosting, unless they're just using another solution which isn't integrated into vexim. But in that case, they would probably be better just commenting out that functionality in vexim, or integrating it with their solution instead. IMO, disabled checks should generally be an exception (e.g. on honeypot mailboxes), not common practice.
That's just food for thought, this can wait till 3.0.

Udera:
As you said, there might be some exceptions that could justify to turn 
off spamassassin (wrongly trained spam-filter, honeypot, accounts for 
automatically processed mails, ...).

Concerning virus mails, I'm not really happy with the current situation 
(just throwing mails away without further notice). Personally, I reject 
all virus mails. If you accept it, you should inform the user from my 
point of view. Either notify that a mail of subject xy from 
xy@example.org was deleted or just strip of the malicious attachment.

We need to get this 2.3-release done in order to get to version 3.0 
soon...

rimas-kudelis:
> We need to get this 2.3-release done in order to get to version 3.0 soon...

yes, certainly. That's why I think further policy changes can wait.

Udera:
I modified the statement as proposed. I run checks with exim, spamassassin is only considered if turned on by the user AND the domain.

rimas-kudelis:
I haven't tested this, but I trust that you have. Merging in... :)

PR #127, opened 2016-03-12
Disable DNSBL checks for authenticated users

hotstepper13:
Adding manitu dnsbl instead and remove spamchecking for authenticated users

### DIFF
--- a/docs/vexim-acl-check-rcpt.conf
+++ b/docs/vexim-acl-check-rcpt.conf
@@ -52,5 +52,16 @@
 #                          and spamassassin='1'}}}{1} {yes}{no}}
 #          !acl          = spf_rcpt_acl
 
+#
+# To use a DNSBL the best would be to set up a caching dns server for you own
+# otherwise you might get rate limited and your requests are not processed
+# 
+# Also disable spamchecking for authenticated users because dial up networks
+# are often marked as bad in case of sending mails
+#
+  accept
+          authenticated = *
+          condition = ${if eq{$authenticated_id}{$sender_address}{yes}{no}}
+
   deny    message       = DNSBL listed at $dnslist_domain\n$dnslist_text
-          dnslists      = zen.spamhaus.org:list.dsbl.org
+          dnslists      = zen.spamhaus.org:ix.dnsbl.manitu.net


Udera:
:+1:

rimas-kudelis:
I had a situation in the past where my server sent out spam because somebody's password wasn't strong enough or was leaked/hacked/sniffed AND SpamAssassin checks were disabled for authenticated users.

Which is why I think that checking for spam is a good idea even if the sender is an authenticated user, so I'm not fond of disabling it.

We could discuss this further if you like. In that case, I suggest to split this patch into two, because I certainly don't mind replacing the retired DNSBL with something else.

By the way, I can't resolve ix.dnsbl.manitu.net. Is this a public service? Does it have a homepage?

Udera:
http://www.dnsbl.manitu.net/?language=en

I also prefer checking all mails by spamassassin. In most cases, you also handle mails of your webserver.

hotstepper13:
In my case the Problem was a false positive for my own dial up ip. Here in Germany we use many dynamic ips which means our provider assign us a new ip each time we dial in again (regulary most providers have a 24 Hour forced reconnect). I am using the Provider Unitymedia and the network ranges for this provider are listed at spamhaus. So if I send a mail to my server it was rejected as the network of my provider was listed at spamhaus. That is the reason why I disabled the dnsbl checking for authenticated users. The patch doesnt disable spamassassin at all. Only the dnsbl checking should be skipped for authenticated users to prevent false positives.

hotstepper13:
Additional info:
I installed also fail2ban in order to observe the mail logfiles. So if someone has 3 authentication errors within one hour he will "banned" by iptables for an hour. This works pretty well and reduce the "load" of the mail server at his have not to handle brute force requests. Luckily my server is not attractive enough to raise efforts (like using a whole botnet with multiple ips for brute forcing).

If you want, i could add installation documentation, too (just tell me if I should use the wiki or documentation within the project)

rimas-kudelis:
@Udera: indeed, if this doesn't affect Spamassassin, then perhaps it's OK?

Udera:
@hotstepper13 The exim-wiki gives a few samples how blacklist-checks can be combined with others to avoid such problems:
https://github.com/Exim/exim/wiki/Verification#sender-verify-failed-and-blacklisted

Blacklists are as well integrated in spamassassin (and you can also ad the ix-manitu-blacklist to it) so if we checked all incoming mails for spam this would be sufficient. Using a list as only rejecting criteria can cause false positives (like your authenticated users).

I see these two options:
* Use blacklist together with a second criteria (sender verify failed, or authenticated)
* Disable blacklist checking and refuse mails above certain spam-score (perhaps: + rDNS checks for non-authenticated users: https://github.com/vexim/vexim2/pull/129)

docs/vexim-acl-check-rcpt.conf
```diff
@@ -52,5 +52,16 @@
 #                          and spamassassin='1'}}}{1} {yes}{no}}
 #          !acl          = spf_rcpt_acl
 
+#
+# To use a DNSBL the best would be to set up a caching dns server for you own
+# otherwise you might get rate limited and your requests are not processed
+# 
+# Also disable spamchecking for authenticated users because dial up networks
```
rimas-kudelis: Can you reword this so that it doesn't sound like we're actually disabling spam checks here, but just DNSBL checks?

Also, I'm unsure if the comment above is even necessary, but even if you want it to stay, I don't think the second one should start with the word "Also".

docs/vexim-acl-check-rcpt.conf
```diff
@@ -52,5 +52,16 @@
...
+# are often marked as bad in case of sending mails
+#
+  accept
+          authenticated = *
+          condition = ${if eq{$authenticated_id}{$sender_address}{yes}{no}}
```
Udera: Just thinking about it, I have a main address with login (`$authenticated_id`) but I often use a different identities with a specific `$sender_address` (redirection to my main account). This wouldn't work.

runout-at:
what about using policyd_weight instead?
config example:  Acl policyd weight #153

Udera:
Should we split this pull request? I think there are no doubts about disabling a non-existing blacklist and we have to put it in the next release.

Then let's open a new topic about the default behavior of vexim. Either we push such things in a vexim 2.3.1 or we directly move to vexim3.

runout-at:
DNSBL config is not the main purpose of vexim but i like the idea of having the examples here.
Which DNSBLs are used should be a decision of the site admin.

We could have the example but should state something like: .."using specific and working DNSBLs is the responsibility of the admin. Things can change and a DNSBL could disappear and/or deliver positive responses on all domains."

Udera:
> DNSBL config is not the main purpose of vexim but i like the idea of having the examples here.
> Which DNSBLs are used should be a decision of the site admin.

sure, but I would keep the current default settings. Comment new or alternative versions unless we decided that this is a much better default behavior for everybody (should then also be on some sort of release notes).

Udera:
This authentication is something perhaps to improve.
The DNSBL-list needed to be updated for version 2.3, so I created a separate PR: #166 
Changing milestone to 2.3.1

docs/vexim-acl-check-rcpt.conf
rimas-kudelis: If we drop the second condition and only leave the `authenticated = *` part, this will work as intended, right?

Udera: Yes it looks like it. Would be good to test it though 😄

Udera: Can't we just use DNSBLs on non authenticated users?
````
deny    message       = DNSBL listed at $dnslist_domain\n$dnslist_text
           !authenticated = *
           dnslists      = zen.spamhaus.org:ix.dnsbl.manitu.net
````

rimas-kudelis: If I understand this correctly, it depends on whether or not we want to keep checking authenticated users. `accept authenticated=*` will mean that this ACL has passed with an accept, and no further checks (DNSBL or any other) will be performed. Merging that condition into the DNSBL block will mean that either the user will be denied (if not authenticated and in DNSBL), or checks will proceed further. Am I correct?

Udera: Yes, currently the check_rcpt-section is much longer and this file is only a small part:
https://github.com/vexim/vexim2/blob/master/docs/configure#L339-L419

We could move the include-command for this file behind line 364 (mails to postmaster are accepted even if the host is on a blacklist, and invalid local-parts are very fast compared to DNS lookups). Then we can move the accept-authenticated-rule (line 414) into this file. And remove lines 369-381.

Should I prepare a PR?

rimas-kudelis: Yeah, I think our config files certainly need more polishing.

rimas-kudelis: Anyway, can you put an `accept   authenticated = *` line as well in one of these files?

rimas-kudelis: Or to just disable DNSBL checks for them, like you suggested.

Udera:
I put the changes into a new PR: https://github.com/vexim/vexim2/pull/205

Udera:
Authenticated users are now accepted regardless of the DNSBL status: https://github.com/vexim/vexim2/pull/205
The old DNSBL was removed here: https://github.com/vexim/vexim2/pull/166

PR #129, opened 2016-03-12
Add reverse lookup

Udera:
Only accept mail from servers with correct reverse DNS settings.

A correct rDNS is required as major mail servers reject mails from servers without one (e.g. gmail).

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -91,6 +91,17 @@ VEXIM_LOCALPART_SUFFIX = +*
 # 'headers_remove' behavior. See https://github.com/vexim/vexim2/pull/102 .
 #OLD_HEADERS_REMOVE = yes
 
+# Reverse DNS checks can be efficient to fight spam. It checks if host name associated to the ip address matches
+# the sender_host_name (for properly configured mail servers this should be the case). However, there are some
+# downsides: DNS checks slow down the delivery process, DNS problems can result in rejected mails. You should
+# consider using a caching DNS server such as unbound on your host.
+# If the rule is activated, failed rDNS entries will only be logged to your exim-logfile (test mode). If you want
+# to reject mails directly, go to the vexim-acl-check-rcpt.conf and replace this line:
+#  warn message           = Warning - Reverse DNS lookup failed for host $sender_host_address.
+# by:
+#  deny message           = Warning - Reverse DNS lookup failed for host $sender_host_address.
+#CHECK_RCPT_REVERSE_DNS = yes
+
 #Debian: hide mysql_servers = localhost::(/var/run/mysqld/mysqld.sock)/vexim/vexim/CHANGE
 #hide mysql_servers = localhost::(/tmp/mysql.sock)/vexim/vexim/CHANGE
 #hide pgsql_servers = localhost/vexim/vexim/CHANGE
--- a/docs/vexim-acl-check-rcpt.conf
+++ b/docs/vexim-acl-check-rcpt.conf
@@ -51,9 +51,12 @@
 #                          where domain = '${quote_mysql:$domain}' \
 #                          and spamassassin='1'}}}{1} {yes}{no}}
 #          !acl          = spf_rcpt_acl
-  deny message           = Warning - Reverse DNS lookup failed for host $sender_host_address.
+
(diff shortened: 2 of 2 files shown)


rimas-kudelis:
Couple questions:

1. why is this being defined here and not in `vexim-acl-check-helo.conf`?
2. let's say this is my outgoing SMTP server, and I don't have an rDNS, and I'm trying to send my e-mail through it. Won't it block me in that case?

rimas-kudelis:
Or is it taken care of by `!authenticated  = *` line, which also explains point point 1 (a user cannot be authenticated yet during the HELO phase)?

Udera:
> Or is it taken care of by !authenticated = * line, which also explains point point 1 (a user cannot be authenticated yet during the HELO phase)?

I don't remember where I found it. But it's possible that


> Or is it taken care of by !authenticated = * line, which also explains point point 1 (a user cannot be authenticated yet during the HELO phase)?



> 2. let's say this is my outgoing SMTP server, and I don't have an rDNS, and I'm trying to send my e-mail through it. Won't it block me in that case?

You will have problems to send mails to most of the public hosts. Or do you mean that you have your local outgoing smtp that uses this server as relay?

Udera:
There are several entries in the exim-wiki concerning different verification processes:
https://github.com/Exim/exim/wiki/Verification

e.g. coupling the rDNS-check with black-lists.

soujak:
I don't think enabling the lookup by default is a good idea, since many administrators may prefer to have a quicker mail processing. Maybe the snippet could be commented out and explained, in order to allow optional activation?

Consider also that in Debian's default configuration there already is a `warn` clause performing DNS verification during RCPT.
```
  .ifdef CHECK_RCPT_REVERSE_DNS
  warn
    condition = ${if and{{def:sender_host_address}{!def:sender_host_name}}\
                      {yes}{no}}
    add_header = X-Host-Lookup-Failed: Reverse DNS lookup failed for $sender_host_address (${if eq{$host_lookup_failed}{1}{failed}{deferred}})
  .endif
```
We may consider:
1. explaining the potential conflict with default configuration;
2. providing a `deny` clause which check the header, instead of (re-)doing the lookup.

Udera:
In my setup, this is the most efficient spam prevention with nearly no 
false-positives (except they have a badly configured mailserver). I'm 
far away from any technical limits so I can't really judge on the 
performance point of view. Considering that you are using a local dns 
cache, recurrent mail hosts will be stored in your cache and the query 
won't take long. If you don't use it and all this junk goes through your 
spam-filter, I'm not sure that this will improve the overall 
performance.

For smaller setups, this might be a helpful option to reduce spam. 
Administrators of larger setups are more experienced and will have to 
change a lot of other settings anyway. So it's perhaps easier to expect 
them to comment this option out.

rimas-kudelis:
I have mixed feelings about this. On the one hand, it's an effective spam prevention technique, but on the other hand, I don't want to ignore cases where people have not or can not set an rDNS record for their servers. Think of a small company running their mail server locally, they won't always be able to get a proper rDNS record for their IP address. Do we really want to cut them off just because of that? I tend to think that policies like this might be more of interest for bigger setups because they naturally will have much more e-mails to process.
I'm not so sure we should just endorse and copy whatever Google, Yahoo or Microsoft are doing, even if not doing that is just like blowing against the wind in the end.

Udera:
Most ipv4-addresses have a reverse-dns-entry. You just need to set your hostname to it if you can't change it. For ipv6 (only or dual stack), this can be a problem.

If you have doubts, we can comment this option out. As the anti-spam options depends on personal preferences, we should provide a section in the Wiki with possible settings and their consequences. I mean that's the advantage of your own mail server.

Udera:
Would it be ok, if I just comment it out? I will then add a part in the wiki, what you can do to fight spam with the pro and cons of each option.

rimas-kudelis:
Or you can `ifdef` it and only comment out the `define`.

Udera:
@soujak @rimas-kudelis 
I updated my pull request with your remarks and disabled the feature by default.

Udera:
Let's get this in for testing.

PR #130, opened 2016-03-19
Bring back message limit for calls to SpamAssassin

soujak:
Re-enable by default the 'accept' ACL clause which skips spam-detection scanning via SpamAssassin for large messages (300 KB). Email scan in this case has a considerable risk to be wastefully killed by scanning timeout, since processing time grows exponentially w.r.t. size, and very little probability of being successful, since most of spam is under 10 KB.

The setting was disabled in commit f8c2c6dc7f982e5d2142586d38c1d6c002f2f2d0 and merged in pull request #24, where a thorough discussion on the subject can be found.


rimas-kudelis:
Can you improve the comment before merge please?

soujak:
:heavy_check_mark:

Issue #133, opened 2016-04-03
quota does not work


runout-at:
i think quota should be given in Mbyte in
/docs/debian-conf.d/transport/30_vexim_virtual_delivery

quota = ${extract{quota}{$address_data}}
should be:
quota = ${extract{quota}{$address_data}}M

Udera:
Exim4U uses this:
````
quota = ${extract{quota}{$address_data}{${value}M}}
````
https://github.com/Exim4U/src/blob/master/etc/exim/exim.conf#L1305

@runout-at : Can you create a pull request and test it?

runout-at:
already tested on my mailserver

Udera:
fixed by https://github.com/vexim/vexim2/pull/149
thanks @runout-at

Issue #135, opened 2016-04-04
domainguess, regex


runout-at:
Would be nice to have a config variable to strip subdomains for domainguess like:
```php
$domainguess_lefttrim = "mail|vexim";
```
for usage in index.php (line 20) 
```php
if($domainguess===1) echo '@'.preg_replace ("/^".$domainguess_lefttrim."\./", "", $_SERVER["SERVER_NAME"]);
```
and login.php (line 14)
```php
if($domainguess === 1 && $_POST['username']!=='siteadmin') $_POST['username'].='@'.preg_replace ("/^".$domainguess_lefttrim."\./", "", $_SERVER["SERVER_NAME"]);
```

Udera:
Is this tested? From my POV, we could include this if you create a pull request. @rimas-kudelis what do you think?

rimas-kudelis:
Fixed by #151

PR #136, opened 2016-04-04
domainguess, host fields, quota bug

runout-at:
solves #135 

cut off string given in $domainguess_lefttrim from left side of hostname


runout-at:
this also closes #133 and sets the quota in Mbyte in transport local_delivery

runout-at:
already tested on my mailserver

runout-at:
The server/service could retrieve a 'host' field from the database to decide if the domain belongs to it.
I suggest to have different 'host' fields for different services: smtp, imap, pop

Dovecot has such a feature for its proxy: http://wiki2.dovecot.org/PasswordDatabase/ExtraFields/Proxy

Database must be altered in tables 'domains' and 'domainalias'. Make sure the new field are populated with 'mail' this is the default host exim would look for with the new config for doing virtual (vexim router) delivery.
With this change it is possible to have several exim servers using the same database for serving different domains.

Add exim config file for mailman which honors the new host fields

runout-at:
this closes #145 'host' fields for services

Issue #138, opened 2016-04-16
acl-check-helo


Udera:
Here we have a problem:
https://github.com/vexim/vexim2/blob/master/docs/vexim-acl-check-helo.conf#L8-L11

I get this error in my logs:
```
H=(166-130.viaservers-b.com.br) [189.1.166.130] temporarily rejected EHLO or HELO 166-130.viaservers-b.com.br: failed to expand ACL string "${if and {{match{$sender_helo_name}{^[0-9]\.[0-9]\.[0-9]\.[0-9]}}{match_ip{$sender_helo_name}{@[]}}}{yes}{no}}": "166-130.viaservers-b.com.br" is not an IP address inside "and{...}" condition
```

rimas-kudelis:
Interestingly, Exim spec says the second condition shouldn't have been evaluated:

> The sub-conditions are evaluated from left to right. The condition is true if all of the subconditions are true. If there are several “match” sub-conditions, the values of the numeric variables afterwards are taken from the last one. When a false sub-condition is found, the following ones are parsed but not evaluated.

By the way, it seems to me like the first condition would only match HELO's consisting of single-digit IP's like 8.8.8.8, or HELO's which start with such IP's (such as 8.8.8.8-foobar-whatever.com). That looks like a bug to me.

Udera:
Just have a look at the condition below, there we are already checking for hostnames beginning with an IP address: https://github.com/vexim/vexim2/blob/master/docs/vexim-acl-check-helo.conf#L14-L16

Therefore, I only kept the check if the incoming server is using our own IP: https://github.com/vexim/vexim2/pull/139

rimas-kudelis:
I think these idea is that either you will comment one of these conditions out, or you'll get different error messages for the case when someone sends your IP as HELO as opposed to sending someone else's (or their own) IP.

And both of these conditions seem broken by matching only single-digit IP addresses (unless that regex is not the flavor I'm familiar with.

I've added a comment on your PR.

Udera:
you could use `match{$sender_helo_name}{[10.0.1.1]}` but I didn't find a way to use `@[]`. Even with other constructions like `forany` to extract the elements of `@[]` didn't work.
From https://github.com/Exim/exim/wiki/AclHeloTricks there is an example:
```
drop    message     = Bad helo name
        condition   = ${if  \
                         and{    \
                             {isip {$sender_helo_name}}  \
                             {match_ip{$sender_helo_name}{@[]}}  \
[...]

PR #139, opened 2016-04-18
Fixing issue #138

Udera:
Fixing https://github.com/vexim/vexim2/issues/138

### DIFF
--- a/docs/vexim-acl-check-helo.conf
+++ b/docs/vexim-acl-check-helo.conf
@@ -4,13 +4,18 @@
   accept hosts = :
   accept hosts = +relay_from_hosts
 
-# The following ACL drops all messages where the HELO argument is *our* IP address
-  drop condition = ${if and {\
-                             {match{$sender_helo_name}{^[0-9]\.[0-9]\.[0-9]\.[0-9]}}\
-                             {match_ip{$sender_helo_name}{@[]}}\
-                            }{yes}{no}}
-       message   = "Dropped spammer pretending to be us"
-# The following ACL drops all messages where the HELO argument is *any* IP address
-  drop condition = ${if match{$sender_helo_name}{^[0-9]\.[0-9]\.[0-9]\.[0-9]}{yes}{no} }
-       message   = "Dropped IP-only or IP-starting helo"
-  accept
+# DROP all messages with raw IP address as HELO (not allowed, it must be enclosed in [])
+drop
+    condition   = ${if isip{$sender_helo_name}}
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
Won't this fail miserably if the HELO identifier is not numeric, just like you reported in #138?

Udera:
yes, that's not working either

Udera:
Could we just use `match` instead of `matchip`?

Udera:
I changed the HELO-ACLs to check these conditions:

- raw IP addresses are not RFC compliant (must be inside [])
- IP address and hostname from server is not allowed as HELO-name

I didn't add further checks like:

- check for an IP part like in `134-24-233-13.dynamic-range.example` because this could block legitimate users, dynamic-ips are as well checked by spamassassin)
- HELO contains my domains or subdomains. You perhaps use your own domain for some hosts which want to send mails and are not part of the relay hosts.
- some more checks on the syntax (single word, point at the end, ...)

However, we could keep some more rules in that are commented by default.

Udera:
Seems to work as it should:
````
2016-07-21 14:29:05 H=(192.168.0.91) [203.171.31.60] rejected EHLO or HELO 192.168.0.91: Access denied - Invalid HELO name (See RFC2821 4.1.3)
2016-07-21 14:39:50 H=(113.162.6.35) [113.162.6.35] rejected EHLO or HELO 113.162.6.35: Access denied - Invalid HELO name (See RFC2821 4.1.3)
2016-07-21 16:39:33 H=(144.76.86.27) [80.78.76.185] rejected EHLO or HELO 144.76.86.27: Access denied - Invalid HELO name (See RFC2821 4.1.3)
2016-07-21 17:51:24 H=rrcs-69-75-209-180.west.biz.rr.com (192.168.0.91) [69.75.209.180] rejected EHLO or HELO 192.168.0.91: Access denied - Invalid HELO name (See RFC2821 4.1.3)
````
The last line is a bit confusing at first. The sender actually used `192.168.0.91` and because I use rDNS, exim4 automatically resolved the hostname associated to `69.75.209.180` (real sender_ip).

In contrast a normal mail with properly configured hostname and DNS entry looks like:
````
2016-07-21 20:51:01 1bQKvq-00015q-Fz <= redacted@hotmail.fr H=bay004-omc2s20.hotmail.com [65.54.190.95] P=esmtps X=TLS1.2:ECDHE_RSA_AES_256_CBC_SHA384:256 S=                                                                                                11033 id=VI1PR04@VI1PR04MB1458.abc.outlook.com 
````

Udera:
We are not changing the default behavior, this PR seems to be working....

PR #142, opened 2016-05-05
Add domains for which a valid DKIM is required

Udera:
Adds a list of domains which are expected to always send mails with a valid DKIM signature. All mails without a valid DKIM signature will be refused. This option is turned off by default.

This can be useful to prevent phishing mails like from amazon, ebay, twitter, ... where mails are always signed with DKIM and no other servers handle their messages. For this reason, this doesn't apply for domains with normal user mails (like gmail, yahoo, hotmail, ...) because of redirected mail or mailing lists.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -87,6 +87,11 @@ hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
 # Comment this line if you want to disable it, instead of + you can use a different separator.
 VEXIM_LOCALPART_SUFFIX = +*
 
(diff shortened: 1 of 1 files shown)


Udera:
I see some problems when several (valid) DKIM signatures are used. We need to figure this out before merging this PR.

rimas-kudelis:
I don't think maintaining a local list of domains like this is feasible in the long run. I'd say it would be an anti-pattern even. I doubt we should promote it.

@Udera, do you think this patch is still relevant?

Issue #109, opened 2016-02-18
vexim-acl-check-helo.conf in split Debian configuration?


ConneXNL:
I was wondering if it's intended that vexim-acl-check-helo.conf is never used in the split debian configuration?

Udera:
I can't see it in the normal debian-settings. @rimas-kudelis created them in the first place, so he should know best ;-)

rimas-kudelis:
I think my rationale for not including these ACL's in Debian setup was that it has nothing to do with virtual e-mail hosting, and I wanted to cover just that in our Debian files. I suppose we can discuss whether or not it should stay that way.

Udera:
In this case all vexim-acl-* settings are not necessary. I would rather provide a complete set of configuration files that work and have reasonable standard settings.

rimas-kudelis:
Well, my idea was that to debianize Vexim truly at some point, and to add a script which would generate a deb package. And I don't think it would be a good practice for a package to ship arbitrary config files, which aren't strictly related to the package functionality (unless they're shipped in `/usr/share/doc/PACKAGENAME/examples/`). But I guess real debianization won't happen until 3.0 or later, so we can just add those files to `conf.d` without much worrying, if that's all it takes.

Udera:
@ConneXNL I tried to fix it here https://github.com/vexim/vexim2/pull/143
I couldn't test it.

rimas-kudelis:
Somewhat relevant: I've filed https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=838098 asking to make ACL configuration more flexible and pluggable.

PR #143, opened 2016-05-05
Add ACL check_helo into debianic config file

Udera:
I hope this fixes https://github.com/vexim/vexim2/issues/109

I had to add a debian-conf.d/acl/30_exim4-config_check_helo myself because the default debian configuration doesn't define any acl_check_helo (in contrast to acl_check_rcpt or acl_check_data).

I didn't have a spare system to test it ;-(

### DIFF
--- a/docs/debian-conf.d/main/00_vexim_listmacrosdefs
+++ b/docs/debian-conf.d/main/00_vexim_listmacrosdefs
@@ -49,5 +49,14 @@ add_environment = MAIN_ADD_ENVIRONMENT
 # Comment this line if you want to disable it, instead of + you can use a different separator.
 VEXIM_LOCALPART_SUFFIX = +*
 
-CHECK_RCPT_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-rcpt.conf
-CHECK_DATA_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-content.conf
+# USE the vexim ACL-rules:
+
+.ifndef MAIN_ACL_CHECK_HELO
+MAIN_ACL_CHECK_HELO = acl_vexim_check_helo
+.endif
+acl_smtp_helo = MAIN_ACL_CHECK_HELO
+
+.ifndef MAIN_ACL_CHECK_DATA
+MAIN_ACL_CHECK_DATA = acl_vexim_check_content
+.endif
+acl_smtp_data = MAIN_ACL_CHECK_DATA
--- a/docs/debian-conf.d/acl/30_vexim_check_helo
+++ b/docs/debian-conf.d/acl/30_vexim_check_helo
@@ -1,0 +1,23 @@
+acl_vexim_check_helo:
+
+  # These ACL's allow filtering by the HELO argument.
+  # The two ACL's overlap. You may want to comment out one of them.
+
+    accept hosts = :
+    accept hosts = +relay_from_hosts
+
+  # DROP all messages with raw IP address as HELO (not allowed, it must be enclosed in [])
+  drop
+    condition   = ${if isip{$sender_helo_name}}
+    message     = Access denied - Invalid HELO name (See RFC2821 4.1.3)
+
+  # Drop all messages where the HELO is _our_ IP address
+  drop condition = ${if eq{[$interface_address]}{$sender_helo_name}}
+     message   = $interface_address is _my_ address
+
+  # DROP all messages with same hostname as *ours*
+  drop  message   = "REJECTED - Bad HELO - Host impersonating [$sender_helo_name]"
+      condition = ${if match{$sender_helo_name}{$primary_hostname}}
+
+  # Accept all others
+  accept
--- a/docs/debian-conf.d/acl/40_vexim_check_data
+++ b/docs/debian-conf.d/acl/40_vexim_check_data
@@ -1,0 +1,48 @@
+acl_vexim_check_content:
+  
+  # First unpack MIME containers and reject serious errors.
+  deny  message		= This message contains a MIME error ($demime_reason)
+        demime		= *
+        condition	= ${if >{$demime_errorlevel}{2}{1}{0}}
+        
+  # Reject typically wormish file extensions. There is almost no
+  # sense in sending such files by email.
+  # Note that you might want to change the list of extensions.
+  deny  message = This domain has a policy of not accepting certain types of attachments \
+                  in mail as they may contain a virus.  This mail has a file with a .$found_extension \
+                  attachment and is not accepted.  If you have a legitimate need to send \
+                  this particular attachment, send it in a compressed archive, and it will \
+                  then be forwarded to the recipient.
+        demime = ade:adep:adp:bas:bat:chm:cmd:cnf:com:cpl:crt:dll:hlp:hta:inf:ins:isp:js:jse:lnk:mad:maf:mag:mam:maq:mar:mas:mat:mav:maw:ocx:pcd:pif:reg:scf:scr:sct:vbe:vbs:wsc:wsf:wsh:url:xnk
+
+  # Reject virus infested messages.
+  warn  message		= This message contains malware ($malware_name)
+        malware		= *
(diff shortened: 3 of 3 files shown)


rimas-kudelis:
I haven't gone through any code, but our files should be called `*_vexim_*`. `*_exim4-config_*` files should only belong to the exim4-config package.

Udera:
Right, I just copied the debian naming scheme from check_rcpt to check_helo, I didn't fix the naming in the end.

docs/debian-conf.d/main/00_vexim_listmacrosdefs
```diff
@@ -42,3 +42,9 @@ VEXIM_LOCALPART_SUFFIX = +*
 
 CHECK_RCPT_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-rcpt.conf
 CHECK_DATA_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-content.conf
+CHECK_HELO_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-helo.conf
```
rimas-kudelis: Is there any reason to want these files outside the usual hierarchy and include them as proposed in this patch?

runout-at: Maybe this is because they are not parsed directly by the debian magic but included by parsed config files. And it might be confusing if they appear in the same directories. - just a guess.

but on the `acl_check_helo:` i think we don't need a separate .conf file.

Udera: I just placed `vexim-acl-check-helo.conf` where the other `vexim-acl-check-*.conf` have been placed. For debian it would make more sense to place them into the `acl`-folder. But then we would have duplicates (or can we use links in github?).

I used a separate file because we did the same in the other configuration file.

runout-at:
The item @[] matches any of the local host’s interface addresses. (exim doku)
What about if the mail host is behind a firewall and has only a local address (like my setup)?
Do a DNS lookup to get the external IP for that mail host?
I would not like to hardcode the IP because if the IP changes i'm sure not to remember to change it here too.

Udera:
There is even a new version on its way:
https://github.com/vexim/vexim2/pull/139/files

But `$interface_address` won't change this either. So you are behind a firewall and the IP could be changing. I don't know how exim could figure this out by itself. The best would probably be to have a script that updates the external IP address inside a textfile. Then exim can read this textfile.

runout-at:
In debian there is in /etc/exim4/update-exim4.conf.conf:
dc_local_interfaces=...
Is it this where @[] gets the IPs from?
Then it might be ok to put the IP there. Its a central point of configuration

...
I think a DNS lookup on the targeted domain should normally do it. But if there are more mail servers behind a firewall this could be more complicated. They might communicate on local IPs and resolve their hostnames internally to this local network.

Maybe this is a very seldom use case.
The ACL would just not refuse emails in this case. So it doesn't hurt too much, but this should be mentioned in the comment of the rule.

runout-at:
about #139 
is it enough to match the sender_helo_hostname against the primary_hostname?
if mail comes for a virtual domain, should sender_helo_hostname be compared against the hostname from the domains MX entry too?

Udera:
> is it enough to match the sender_helo_hostname against the primary_hostname?
> if mail comes for a virtual domain, should sender_helo_hostname be compared against the hostname from the domains MX entry too?

In general, your hostname should be unique (you could have mail1.example.org, mail2.example.org, ...). So there can be several mx-entries. If you use only one mailserver, you could also include `VIRTUAL_DOMAINS` and `ALIAS_DOMAINS`

> Is it this where @[] gets the IPs from?
> Then it might be ok to put the IP there. Its a central point of configuration

I don't know. Before you start a difficult configuration. I have run the settings of #139 for some time, and the most efficient rule was the one blocking IP addresses:
````
drop
    condition   = ${if isip{$sender_helo_name}}
    message     = Access denied - Invalid HELO name (See RFC2821 4.1.3)
````
Most hosts who tried to use my ip address, forgot that the correct helo is `[1.2.3.4]` and not `1.2.3.4`. So the own-ip-rule was rarely needed.
If you want to know the impact, you can just put manually your current ip address in for a day and check your logfile.

File: docs/debian-conf.d/main/00_vexim_listmacrosdefs (hunk 1 above)
Comment (rimas-kudelis) [inline docs/debian-conf.d/main/00_vexim_listmacrosdefs]: I think in Debian we should just copy all of these files under `acl/` and give them names with `_vexim_`. These macros are meant for private use, there's no need for us to usurp them.

**Comment by Udera** [comment]
Like this? The acl_check_rcpt of vexim would need a slightly larger acl-file:
https://github.com/vexim/vexim2/blob/master/docs/configure#L332-L360

The blacklists are as well provided by the acl_check_rcpt from debian.

File: docs/debian-conf.d/main/00_vexim_listmacrosdefs [hunk 2]
```diff
@@ -49,5 +49,14 @@ add_environment = MAIN_ADD_ENVIRONMENT
...
+
+.ifndef MAIN_ACL_CHECK_HELO
+MAIN_ACL_CHECK_HELO = acl_vexim_check_helo
+.endif
+acl_smtp_helo = MAIN_ACL_CHECK_HELO
+
+.ifndef MAIN_ACL_CHECK_DATA
+MAIN_ACL_CHECK_DATA = acl_vexim_check_content
+.endif
+acl_smtp_data = MAIN_ACL_CHECK_DATA
```
rimas-kudelis: The second block seems to already exist in `main/02_exim4-config_options`.
Perhaps we should omit it and add the first one into `main/02_vexim_options` ?

rimas-kudelis:
Yes, like this. But did you forget to add vexim_check_rcpt?
Also (this could be done in a followup) I think we should check that we aren't shipping duplicate stuff with the `exim4-config` package.

rimas-kudelis:
You know, I'm afraid this won't work after all. From my understanding, only one ACL can be used in a particular scenario, and if we define our own ACL as the one to use, this means we're overriding the default ACL that comes with `exim4-config`. I don't think we want that, do we?

Considering we don't want to spend too much time and delay our release even further just to figure out how to do this properly, I guess we should go back to the previous (current) approach of defining `CHECK_*_LOCAL_ACL_FILE` macros and pointing them to our files. The *right way* can then wait until 3.0.

For future reference, I think we should look at how other packages solve this problem of integration with Exim. For example, the `greylistd` package ships a `greylistd-setup-exim4` script, which adds the necessary blocks right into the stock config. I wonder if this is actually a good way to do this, I guess it would be best to discuss it with Exim maintainers in Debian.

Udera:
> Yes, like this. But did you forget to add vexim_check_rcpt?

This only adds the blacklists which are already part of the default debian config. No real benefit to override the Debian settings.



> if we define our own ACL as the one to use, this means we're overriding the default ACL that comes with exim4-config. I don't think we want that, do we?

It depends what we are aiming for. Problem is that we don't control the ACL rules and that it might change over time. This is perfect if someone very experienced just wants to use the virtual-delivery routers and transport of vexim and then configures all the ACLs to his personal preference.

Normal users probably prefer a complete configuration where they just replace files and make some minor adjustments.

Let's consider this as part of https://github.com/vexim/vexim2/issues/160 and move it to the future.

rimas-kudelis:
I think we can leave them as macros (as you did previously) in 2.3.

Udera:
I think I deleted the previous version....

rimas-kudelis:
OK, let's leave this for future then.
We've still got to see how @ikonia's package generation venture is doing. :)

Udera:
Closing in favor of: https://github.com/vexim/vexim2/issues/210

rimas-kudelis:
Reopening for better tracking. #210 is more of a meta-issue, I'm not sure if it will end up producing anything in the end.

rimas-kudelis:
Ah, this was a PR and not an issue. Sorry for bugspam.

PR #149, opened 2016-06-14
bugfix quota im MB

runout-at:


rimas-kudelis:
This should be fixed in `docs/configure` as well.

runout-at:
now fixed in `docs/configure` as well.

rimas-kudelis:
thank you!

PR #150, opened 2016-06-14
LMTP transport for dovecot

runout-at:
if dovecot is used for serving imap than it is better to let dovecot do the delivery.
it is done by using the LTMP protocol.

this is just an example exim config which is disabled by default and can be enabled by renaming 2 router-files as described in the header of the lmtp-router-file


Udera:
I currently pipe mails to dovecot in order to benefit from all the dovecot features (sieve):
`| /usr/lib/dovecot/dovecot-lda -d $local_part@$domain`
Your current approach would pass everything to lmtp, and you couldn't decide on a account-base? And neither reply_transport.
````
  file_transport = virtual_delivery
  reply_transport = address_reply
  pipe_transport = address_pipe
````

runout-at:
You're right. i need an additional router if exim should handle that.

In my setup (and it's just an example) exim is the MTA and dovecot the MDA.
dovecot has support for sieve, can pipe to commands,...
dovecot can use the vexim tables, so there is no difference for the postmaster and user.

it would be only a different way to route, transport and delivery of emails.
i think dovecot is the better choice for the delivery.

my plan is, to process vacation

runout-at:
i added the pipe router now.

do we need the reply router? what is the purpose?

Udera:
Needs some testing. I'm suggesting to move that to version 2.3.1.

Udera:
Wouldn't it be better, to add an option to use lmtp for mail deliveries (checkbox)? And then query the database and set 
`transport = virtual_delivery_lmtp`
instead of
`transport = virtual_delivery` ?

runout-at:
I was thinking about a server wide setting.
If we will do it per domain or per user it needs a lot of work to integrate into the database, frontend,...
Maybe a feature for a later version.
Right now i would commit it into master just as an example.

Udera:
> If we will do it per domain or per user it needs a lot of work to integrate into the database, frontend,...

It's a small option, not so much work to do (siteadmin-page, useradd, adminuseradd).

For a temporary workaround, I would prefer a more compact solution:

* Add transport `virtual_delivery_lmtp`
* then activate lmtp with a variable and the .ifdef-construction, if lmtp is enable use:
`file_transport = virtual_delivery_lmtp` else `file_transport = virtual_delivery`.

That makes handling of the configuration files much easier. And later we can easily fetch this setting from the database instead of this manual selector.

rimas-kudelis:
Agree with Udera. Also, instead of moving and adding Dovecot files into the repository, I suggest to edit https://github.com/vexim/vexim2/wiki/Server-configuration:-Dovecot .

PR #151, opened 2016-06-14
Domainguess left trim

runout-at:
enhancement of domainguess feature

solves #135

cut off string given in $domainguess_lefttrim from left side of hostname

### DIFF
--- a/vexim/config/variables.php.example
+++ b/vexim/config/variables.php.example
@@ -40,8 +40,11 @@
 
   /* Guess domain name from hostname and allow login based on
      local part only. It is off by default, set this value to 1
-     in order to enable this function */
+     in order to enable this function 
+     and set a string which will be cut off the left side of
+     the hostname*/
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
From my understanding, with this variable, the regex expands to something like this:
```
/^mail|vexim\./
```
right?
If so, then I think this is not correct, as it matches `^mail` or `vexim\.`. I think you should use braces. around the contents of `$domainguess_lefttrim` variable whne using it in the regex.

runout-at:
ok. you are right. i will fix this.

runout-at:
now it should expand to `/^(mail|vexim)\./`

Udera:
I tested this PR:
mail.example.org -> @example.org
mailn.example.org -> @mailn.example.org
quarks.mail.example.org -> @quarks.mail.example.org

So it does what it should.

PR #152, opened 2016-06-15
Password strength

### LINKED ISSUE #134: check passwort complexity
It would be very useful to check the complexity of a new/changed password
If the _clear_ field disappears we have no chance to check later if people use strong passwords. And i really know... users just don't care... I see weak passwords on mail accounts too often!

I would suggest to include something like this 
http://code.runnable.com/UmrnTejI6Q4_AAIM/how-to-validate-complex-passwords-using-regular-expressions-for-php-and-pcre
in vexim/config/functions.php: function crypt_password

runout-at:
closes #134 `check passwort complexity`

this is the server side implementation.
could be improved by using javascript to tell the user if a password is good before submitting it.

additionally i changed the javascript for generating passwords with 12 characters and added more digits to have a higher chance to have a digit in the password.
i'm not a javascript programmer and i don't know how to ensure to have at least one character of each type in the password.

### DIFF
--- a/vexim/config/functions.php
+++ b/vexim/config/functions.php
@@ -28,8 +28,8 @@
           || preg_match_all('$\S*(?=\S{12,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])\S*$', $candidate)
           || preg_match_all('$\S*(?=\S{16,})(?=\S*[a-z])(?=\S*[A-Z])\S*$', $candidate)
           || (strlen($candidate)>20)
-        )
-            return TRUE;
+        ) { return TRUE; }
+
         return FALSE;
     }
     
--- a/vexim/config/functions.php
+++ b/vexim/config/functions.php
@@ -24,7 +24,8 @@
      * @return  boolean  true if password is strong enough
      */
     function password_strengthcheck($candidate) {
-        if (preg_match_all('$\S*(?=\S{8,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])(?=\S*[\W])\S*$', $candidate)
+        if ( $passwordstrengthcheck == 0
+          || preg_match_all('$\S*(?=\S{8,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])(?=\S*[\W])\S*$', $candidate)
           || preg_match_all('$\S*(?=\S{12,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])\S*$', $candidate)
           || preg_match_all('$\S*(?=\S{16,})(?=\S*[a-z])(?=\S*[A-Z])\S*$', $candidate)
           || (strlen($candidate)>20)
--- a/vexim/config/functions.php
+++ b/vexim/config/functions.php
@@ -35,6 +35,18 @@
         ) { return TRUE; }
 
         return FALSE;
+/*
+    Explaining $\S*(?=\S{8,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])(?=\S*[\W])\S*$
+    $ = beginning of string
+    \S* = any set of characters
+    (?=\S{8,}) = of at least length 8
+    (?=\S*[a-z]) = containing at least one lowercase letter
+    (?=\S*[A-Z]) = and at least one uppercase letter
+    (?=\S*[\d]) = and at least one number
+    (?=\S*[\W]) = and at least a special character (non-word characters)
+    $ = end of the string
+
+ */
     }
     
     /**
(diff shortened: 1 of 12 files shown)


vexim/config/functions.php
```diff
@@ -15,7 +15,40 @@ function validate_password($password, $confirmationPassword)
...
+     */
+    function password_strengthcheck($candidate)
+    {
+        global $passwordstrengthcheck;
+        
+        if ( $passwordstrengthcheck == 0
+          || preg_match_all('$\S*(?=\S{8,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])(?=\S*[\W])\S*$', $candidate)
+          || preg_match_all('$\S*(?=\S{12,})(?=\S*[a-z])(?=\S*[A-Z])(?=\S*[\d])\S*$', $candidate)
+          || preg_match_all('$\S*(?=\S{16,})(?=\S*[a-z])(?=\S*[A-Z])\S*$', $candidate)
+          || (strlen($candidate)>20)
```
rimas-kudelis: 20?????

rimas-kudelis: ah, nevermind, I see the idea now.

rimas-kudelis:
To be honest, I'm not sure we want to enforce "strong" passwords even for the admin user. I tend to think that the administrator should have should have a clue themself, or may have reasons to use a "simpler" password.

The current JS generator function was copied from PhpMyAdmin. It can certainly be improved, but I don't think that just increasing the likelihood of getting a number there is the best approach. There are some other generators available on the net, but the ones I found (e.g. [here](http://stackoverflow.com/questions/1497481/javascript-password-generator)) either have their flaws as well, or are basically same as ours.

Maybe I'm too relaxed about this, but I think it's most important to not have matching usernames and passwords. All other "enforcements" are sometimes just as annoying as they are secure.

@Udera what do you think?

Either way, all commits should be squashed into one before merging.

Udera:
I think it is good to raise the awareness to use secure passwords. However, enforcing them is probably not the best way to do it. Today, even the most shitty forum enforces a strict password policy where my standard password is not strong enough and I need to add more special characters. So what to do? I just add `!!!!`! This human weakness is already used by password hackers and astonishingly good passwords are decrypted: http://arstechnica.com/security/2013/05/how-crackers-make-minced-meat-out-of-your-passwords/

Coming up with a good memorable password is not so easy. Perhaps a better way would be to provide a warning with a link to a guide on how to construct secure passwords? I have no experience if this really helps and if users will pay attention to it.

runout-at:
i came up with this because i have a big problem with some users. and last days i figured out even with postmasters :( that's why i would also force admins to use secure passwords.
for myself i use a password save because i have over 100 passwords.

i can put a compare of the password with the username in the check easily and will do it.

Udera:
I had an interesting experience: Admins run a script (like e.g. https://rundata.wordpress.com/2013/12/29/linux-password-cracker/) that checks every night to hack the accounts, if it succeeded, a mail asked you to change your account password because it was hacked within xx minutes. Due to this experience, I thought about a different more secure password.

Udera:
Are there good reasons not to implement this feature? In principle, it is good to have secure mail passwords. If this is not wanted in some environments, this option can be disabled (e.g. for testing).

I wouldn't change the random-generator. A higher probability for numbers makes the password more predictable.

runout-at:
I changed the random-pw-generator ro raise the probability to have a number in the password. Better would be to force at least one number in the password.

Udera:
> I changed the random-pw-generator ro raise the probability to have a number in the password. Better would be to force at least one number in the password.

Why? Such bias on the results will weaken passwords. And doing the maths (hopefully I'm not wrong), ~14% of the passwords should end up without numbers.

runout-at:
As i said: Better would be to force at least one number in the password.
I think if there is no number in the PW it will weaken the password too. I didn't calculate what's worse ;)

Udera:
There is a random password generator which also uses combinations of characters:
https://www.random.org/passwords/?num=100&len=12&format=html&rnd=new
As you can see, there are some entries without any number. If you enforced a number, an attacker would know that he doesn't have to try all combinations without a number (~10^20 different combinations on a 12 character password)!

Udera:
Can you undo the doubling or tripling of the numbers?
And than squash everything in a single commit: http://stackoverflow.com/questions/8213926/git-simplest-way-of-squashing-commits-on-master#8214049

Udera:
I gave it a quick testing and it seems to do what it should. @rimas-kudelis what do you finally think of it?

rimas-kudelis:
So our frontend will not necessarily generate passwords with numbers, and will never add special characters to them, yet the backend will require them? This doesn't really seem right.

Udera:
If you don't put special characters, you will need a longer password. For that reason, the password length of generated passwords was extended to 12 characters.

rimas-kudelis:
Ah, I see. Alright, let's merge.
Although I'm still not a fan of this. :)

PR #153, opened 2016-06-15
Acl policyd weight

runout-at:
example config for policyd-weight instead of using DNSBL


Udera:
You can add some smtp syntax checking in the acl-part (and SPF, ...). Spamassassin is as well included which also can add checks against block-lists. What is the actual benefit of policyd_weight which can't be handled in exim directly?

runout-at:
I like policyd_weight because i had some troubles with DNSBLs and false positives in the past and polycd_weight does something more complex than just use a singl DNSBL result for blocking a sender.
Its very easy to use with exim additionally to spamassassin.

I'd include the config in vexim as an example. People can decide if they like to enable ist.

Vexim has a different purpose, but there are already some other example configs which do not really relay to virtual users and in that way it makes sense to put some more examples to the repository.
I added this as an alternative to using DNSBLs directly because of the discussion #127

Udera:
policyd was written for postfix which has a more modular design than exim. Many rules (syntax checking during smtp) can directly be implemented in exim. I'm currently using only very conservative DNSBLs and I integrate others into spamassassin. There is perhaps a way, to combine several DNSBLs in exim that mail are only rejected if the host is listed in more than one DNSBL.

runout-at:
It's just an example how to use it if somebody wants it.
It would be very nice if someone takes the time and makes a real exim implementation of the policyd_weight ruleset

Udera:
Let's focus on the v2.3-release and then try to establish some policyd-like behavior.

rimas-kudelis (changes requested):
If this is part of the RCPT ACL, it should probably still be in the `vexim-acl-check-rcpt.conf` file, but either commented out or enclosed inside a `.IFDEF`.

But also, I don't think that just adding an example is sufficient. As a mail admin, I wouldn't know if I want to enable and configure policyd_weight or not. I think we should always either explain very clearly why a specific disabled or commented out example exists in the config file, or not have such examples at all. Otherwise this config just becomes a mess of examples with no clear purpose or guaranteed interoperability and a pain in the butt to maintain.

PR #156, opened 2016-07-14
Allow forwarding to several destinations

### LINKED ISSUE #154: Forward mail to different addresses for pop/imap-account
## How to reproduce
1. Create POP/IMAP account
2. Edit settings (`adminuserchange.php`), forward mails to a list of other mail addresses, e.g. user1@example.org,user2@example.org

## Expected behavior
it should work.
and select box should allow selection of several addresses

## Actual behavior
`user1@example.org,user2@example.org is not a valid e-mail address.`

Udera:
Fix #154 
There is also a small javascript that activates the input field for forwarding when the checkbox for forwarding is active. The problem is when the checkbox is not checked, all inputs to the textfield are ignored.

### DIFF
--- a/vexim/adminuserchange.php
+++ b/vexim/adminuserchange.php
@@ -29,8 +29,25 @@
     <title><?php echo _('Virtual Exim') . ': ' . _('Manage Users'); ?></title>
     <link rel="stylesheet" href="style.css" type="text/css">
     <script src="scripts.js" type="text/javascript"></script>
+    <script type='text/javascript'>
+      function fwac() {
+      document.getElementById('forward').disabled = !document.getElementById('on_forward').checked;
+      document.getElementById('forwardmenu').disabled = !document.getElementById('on_forward').checked;
+      }
+      function boxadd() {
+        var exstring = document.getElementById('forward').value;
+        var box = document.getElementById('forwardmenu');
+        var selectitem = box.options[box.selectedIndex].value;
+        if (!exstring.match(/\S/)) {
+          document.getElementById('forward').value=selectitem;
+        } else {
+          document.getElementById('forward').value += "," + selectitem;
+        }
+      }
+
+    </script>
   </head>
-  <body onLoad="document.userchange.realname.focus()">
+  <body onLoad="document.userchange.realname.focus(); fwac()">
   <?php include dirname(__FILE__) . '/config/header.php'; ?>
     <div id="Menu">
       <a href="adminuser.php"><?php echo _('Manage Accounts'); ?></a><br>
@@ -265,20 +282,20 @@
         </tr>
         <tr>
           <td><?php echo _('Forwarding on'); ?>:</td>
-          <td><input name="on_forward" type="checkbox" <?php
+          <td><input name="on_forward" id="on_forward" type="checkbox" <?php
             if ($row['on_forward'] == "1") {
               print " checked";
-            } ?>>
+            } ?> onchange="fwac()" onclick="fwac()">
           </td>
         </tr>
         <tr>
-          <td><?php echo _('Forward mail to'); ?>:</td>
+          <td valign="top"><?php echo _('Forward mail to'); ?>:</td>
           <td>
-            <input type="text" size="25" name="forward"
+            <input type="text" size="25" name="forward" id="forward"
             value="<?php print $row['forward']; ?>" class="textfield"><br>
-            <?php echo _('Must be a full e-mail address'); ?>!<br>
-            <?php echo _('OR') .":<br>\n"; ?>
-            <select name="forwardmenu">
+            <?php echo _('Enter full e-mail addresses, use commas to separate them'); ?>!<br>
+            <?php echo _('or select from this list') .":<br>\n"; ?>
+            <select name="forwardmenu" id="forwardmenu" onchange="boxadd()">
               <option selected value=""></option>
               <?php
                 $queryuserlist = "SELECT realname, username, user_id, unseen
--- a/vexim/adminuserchangesubmit.php
+++ b/vexim/adminuserchangesubmit.php
@@ -13,6 +13,7 @@
 	  header ("Location: adminuser.php?failupdated={$_POST['localpart']}");
 	  die();  
   }
+  $account = $sth->fetch();
  
   # Fix the boolean values
   $query = "SELECT avscan,spamassassin,pipe,uid,gid,quotas,maxmsgsize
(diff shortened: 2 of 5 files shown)


rimas-kudelis:
> The problem is when the checkbox is not checked, all inputs to the textfield are ignored.

I guess it's a feature, not a problem. :) But surely, disabling the fields when forward is actually off makes things much more obvious.

vexim/adminuserchange.php
```diff
@@ -30,7 +30,13 @@
     <link rel="stylesheet" href="style.css" type="text/css">
     <script src="scripts.js" type="text/javascript"></script>
   </head>
-  <body onLoad="document.userchange.realname.focus()">
+  <script language='JavaScript' type='text/JavaScript'>
```
rimas-kudelis: no need for `language='JavaScript'`, and `text/javascript` should go in lowercase.

vexim/adminuserchangesubmit.php
```diff
@@ -30,10 +30,15 @@
   }
   if (isset($_POST['on_forward'])) {
     $_POST['on_forward'] = 1;
-    if(!filter_var($_POST['forward'], FILTER_VALIDATE_EMAIL)) {
-      header ("Location: adminuser.php?invalidforward=".htmlentities($_POST['forward']));
-      die;
+    $forwardto=explode(",",$_POST['forward']);
```
rimas-kudelis: All of this assumes `$_POST['forward']` is not empty. What if it is?

Also, `$_POST['forwardmenu']` is not even checked for correctness later on, and it overwrites whatever was specified in `$_POST['forward']`. This doesn't seem very nice at all. What would you think about adding some more JS which would overwrite the contents of the `forward` field in the front-end when an address is selected in the drop-down (onchange), and then simply ignoring the contents of `$_POST['forwardmenu']` in the backend altogether? In that case, this assumption above that `$_POST['forward']` must not be empty would be correct.

rimas-kudelis: alternatively the contents of `$_POST['forwardmenu']` could be appended to `$_POST['forward']`, but this would mean changing strings in the UI as well.

rimas-kudelis:
Also, this patch leaves `userchange.php` and `userchangesubmit.php` unfixed.

vexim/adminuserchangesubmit.php
Udera: `$_POST['forwardmenu']` makes everything difficult. I thought about a JS solution, however, if you select one address, it is copied to the field, but what should it do when you select a different one, should it be added to a list, should it be overwritten? The latter would be the easiest to implement and covers probably 90% of all cases. I think, I'll go with that and we implement a more sophisticated solution in the next version.

Udera:
> Also, this patch leaves userchange.php and userchangesubmit.php unfixed.

We have way too much duplicate code, you always miss something ;-(

vexim/adminuserchangesubmit.php
rimas-kudelis: That's what it did until now: overwrote whatever was in the text field with a single chosen email address. Implementing that in the front-end the same fashion won't even change the funtionality, just make it more obviuos.
One thing I'm a bit worried about is that if a user chooses an address accidentally, they can select an empty value again before submitting, thus leaving the contents of the forward field intact. If we make the selection propagate into the text field immediately, it will be possible to mess up the contents of one's `forward` field by accident.
Perhaps adding the address after a comma is a better approach. Alternatively, perhaps if the initial contents of the field aren't empty, there should be a Yes/No dialog?

Udera:
I changed:
* multi-address forward is also possible in `userchange.php`
* selectbox `forwardmenu` is copied with js to `forward`, if `forward` is not empty, the address is added at the end of the list. `forwardmenu` is not processed in php.
* An empty `forward`-field will disable the forward-function as well.

vexim/adminuserchange.php
```diff
@@ -265,20 +282,20 @@
...
+            } ?> onchange="fwac()" onclick="fwac()">
           </td>
         </tr>
         <tr>
           <td><?php echo _('Forward mail to'); ?>:</td>
           <td>
-            <input type="text" size="25" name="forward"
+            <input type="text" size="25" name="forward" id="forward"
             value="<?php print $row['forward']; ?>" class="textfield"><br>
             <?php echo _('Must be a full e-mail address'); ?>!<br>
```
rimas-kudelis: I think we should change these two labels as well.
E.g. "Enter full e-mail addresses, use commas to separate them".
And then something instead of the "OR", but not sure what right now.

Udera:
I updated the labels.

vexim/userchange.php
```diff
@@ -133,12 +138,12 @@
...
-  	    <td><input name="on_forward" type="checkbox"
-          <?php if($row['on_forward'] == "1") { print " checked "; } ?>>
+  	    <td><input name="on_forward" type="checkbox" id="on_forward"
+          <?php if($row['on_forward'] == "1") { print " checked "; } ?> onchange="fwac()" onclick="fwac()">
           </td></tr>
   	  <tr><td><?php echo _("Forward mail to");?>:</td>
-	    <td><input type="text" name="forward" value="<?php print $row['forward']; ?>" class="textfield"><br>
-          <?php echo _("Must be a full e-mail address"); ?>
+	    <td><input type="text" name="forward" id="forward" value="<?php print $row['forward']; ?>" class="textfield"><br>
+          <?php echo _("Must be a full e-mail addresses, use commas to separate them."); ?>
```
rimas-kudelis: Let's make this label identical to that displayed to the admin.

rimas-kudelis:
Thanks. One last nit and we're good to go.
Although we could also ignore empty email addresses (e.g. if `email@address,,email@address2` might not pass due to a double-comma), but I guess this can wait until later.

Udera:
> Although we could also ignore empty email addresses (e.g. if email@address,,email@address2 might not pass due to a double-comma), but I guess this can wait until later.

Yes there we might need some more logic, and perhaps we should then put this in a common function to be called from the script. I put this to the 2.3.1-milestone.

runout-at:
what if you want to disable forwarding temporary?
than you can not edit the list of forward targets if the checkbox is unchecked?

For the target list i would prefer a text box with one mail address per line.

vexim/adminuserchangesubmit.php
```diff
@@ -28,12 +28,17 @@
...
+      $forwardto[$i]=trim($forwardto[$i]);
+      if(!filter_var($forwardto[$i], FILTER_VALIDATE_EMAIL)) {
+        header ("Location: adminalias.php?invalidforward=".htmlentities($forwardto[$i]));
+        die;
+      }
     }
+    $forwardaddr = implode(",",$forwardto);
   } else {
     $_POST['on_forward'] = 0;
     $_POST['forward']='';
```
rimas-kudelis: This should actually set `$forwardaddr` instead.

rimas-kudelis:
> what if you want to disable forwarding temporary?
> than you can not edit the list of forward targets if the checkbox is unchecked?

That's by design. What's not by design (not by mine at least) is that if you uncheck the box, your forward addresses are cleared. I think this should not be the case. @Udera, what do you think?

> For the target list i would prefer a text box with one mail address per line.
It's a good thing to want, but I don't think it's worth implementing in 2.x. Our UI (and our architecture as well) is so outdated it needs a huge overhaul anyway. On the other hand, patches are welcome.

Udera:
> what if you want to disable forwarding temporary?
>     than you can not edit the list of forward targets if the checkbox is unchecked?
> 
> That's by design. What's not by design (not by mine at least) is that if you uncheck the box, your forward addresses are cleared. I think this should not be the case. @Udera, what do you think?

Yes, vexim should keep the settings. When deactivating the forwarding, I will keep the value from the database for two reasons:
* a deactivated field is submitted empty (so we can't read it out)
* the values of the `forward`-field would enter the database unchecked

I made a separate commit that you can see the changes. In the end I can merge all commits into one.

Udera:
Put all in one commit and merged.

PR #159, opened 2016-07-18
check if user is enabled

runout-at:


runout-at:
#158

Udera:
Makes sense to put this in the complete config-file (https://github.com/vexim/vexim2/blob/master/docs/configure) and the imap authentication (e.g. https://github.com/vexim/vexim2/blob/master/docs/clients/dovecot.txt) as well.

runout-at:
for an sql example in dovecot i just added the file to the branch
https://github.com/runout-at/vexim2/blob/lmtp-for-dovecot/docs/clients/dovecot/dovecot-sql.conf.ext

Udera:
Yes, but this is a little bug fix that can easily be tested and then merged for the vexim2.3-release.

runout-at:
ok. done.

PR #161, opened 2016-07-18
spamtrap/honeypot, backscatterer

runout-at:
spamtrap/honeypot:
define a list of email addresses which you lay out as a bait on websites "CONFDIR/spamtrap_receiver"
one address per line
IPs of senders to these addresses will be logged to "/etc/exim4/local_host_blacklist"
and be rejected in the future
just put the bait email addresses in a hidden/invisible area on your website
i gathered ~500 IPs in one month

backscatterer:
as i see lots of mails coming from somwhere but the sender pretends to be my domain i created this acl
mails from "abc123@example.com" to "xyz987@example.com" will be ckecked against the backscatterer DNSBL

### DIFF
--- a/docs/vexim-acl-check-rcpt.conf
+++ b/docs/vexim-acl-check-rcpt.conf
@@ -1,3 +1,18 @@
+# spamtrap write sender IP to BL file when receiving messages to special
+# adresses defined in the file 'CONFDIR/spamtrap_receiver'
+deny
+  message = ${run{/bin/bash -c "/bin/echo $sender_host_address >> /etc/exim4/local_host_blacklist;"}}
+  condition = ${lookup{$local_part@$domain}lsearch{CONFDIR/spamtrap_receiver} {yes}{no}}
+  logwrite = :main,reject: $sender_host_address - IP logged, $local_part@$domain is only used by spammers
+
+# drop backscatterer messages
+deny
+  senders = : $local_part@$domain
+  dnslists  = ips.backscatterer.org
+  add_header = X-blacklisted-at: $dnslist_domain
+  log_message = sender same as recipient $local_part@$domain and $sender_host_address listed at $dnslist_domain
+
+
 # Use spfquery to perform a pair of SPF checks (for details, see
 # http://www.openspf.org/)
 # This check has been copied from a stock Debian config and has not been


Udera:
I take all these spam-trap-mails and forward them to one single account which is then used to train spamassassin. Perhaps we can implement a more centralized solution in the future.

Bounce address tag validation can be used to fight backscatter: https://en.wikipedia.org/wiki/Bounce_Address_Tag_Validation

Udera:
Sorry, I didn't realize that this is actually a pull request.

docs/vexim-acl-check-rcpt.conf
```diff
@@ -1,3 +1,18 @@
+# spamtrap write sender IP to BL file when receiving messages to special
+# adresses defined in the file 'CONFDIR/spamtrap_receiver'
+deny
+  message = ${run{/bin/bash -c "/bin/echo $sender_host_address >> /etc/exim4/local_host_blacklist;"}}
```
Udera: And where is `/etc/exim4/local_host_blacklist` used? If you block mails from hosts in `/etc/exim4/local_host_blacklist` you have to consider that only an account was hijacked (gmail, yahoo, ...) and also a grace period after which you remove the host.

docs/vexim-acl-check-rcpt.conf
```diff
@@ -1,3 +1,18 @@
+# spamtrap write sender IP to BL file when receiving messages to special
+# adresses defined in the file 'CONFDIR/spamtrap_receiver'
+deny
```
Udera: You really want to deny the message? Then it will be removed from the spammers list. I mean you want to receive spam on this spamtrap.

Udera:
Regarding the DNSBL, we should update them urgently. Perhaps you can take over: https://github.com/vexim/vexim2/pull/127/files
And remove the non-existant blacklist. You can just add a further recommendations (backscatter-list, ix...) admins can activate if they want.

runout-at:
i should have done two commits for that ;(

about DNSBL: i don't use it, thats why i will not take #127.
and adding backscatterers there is a problem. you'll get too many false positives. that's why i use it only for senders who pretend to be me.

runout-at:
on the spamtrap/honeypot:
you are right! i was already thinking about a timeout feature. i'd like to do this with `ipset`. just didn't find time yet.
and i have to look into my acl file, maybe i missed something to copy

docs/vexim-acl-check-rcpt.conf
runout-at: /etc/exim4/conf.d/acl/30_exim4-config_check_rcpt
around line 216:

```
  # deny bad sites (IP address)
  # CONFDIR/local_host_blacklist holds a list of host names, IP addresses
  # and networks (CIDR notation)  that should have their access denied to
  # The local host. Messages coming in from a listed host will have all
  # RCPT statements rejected.
  #
  # The explicit white lists are honored as well as negative items in
  # the black list. See exim4-config_files(5) for details.
  deny
    message = sender IP address $sender_host_address is locally blacklisted here. If you think this is wrong, get in touch with postmaster
    !acl = acl_local_deny_exceptions
    hosts = ${if exists{CONFDIR/local_host_blacklist}\
                 {CONFDIR/local_host_blacklist}\
                 {}}
```

Udera:
Is it really worth running your own blacklist? Managing is a bit complicated and will be more efficient if you have a larger network of mail servers. I just got another idea, less perfect but easier to implement and to maintain:

> Drop if any of the recipients mentioned is one which only ever receives spam (ideally, a spam "trap" address):

```
drop    condition = ${lookup{$local_part@$domain}lsearch{/etc/exim/only-used-by-spammers} {yes}{no}}
        logwrite = :main,reject: $sender_host_address - $local_part@$domain is only used by spammers
        message = I don't think so
```
from: https://github.com/Exim/exim/wiki/AclSmtpRcpt

runout-at:
Actually https://github.com/Exim/exim/wiki/AclSmtpRcpt is the source of my idea.

I'm testing a new version now, using ipset and iptables to block blacklisted hosts before they hit exim.
Because ipset has timeouts for IP entries there should not be any management of the blacklist necessary. Just the setup is more complicated because of the integration in the firewall.
To block connections on the firewall seems to be evil but they get not blocked forever (testing with 5h). 
I did move most of these rules to the helo-acl.

After running the 'new' version for some days, i get just ~10% of the bad connections logged to the ipset blacklist.

Udera:
I luckily don't have much of such mails, so it is difficult for me to test. For me the problem about these blacklists is that you need to keep an eye on them, you probably need to whitelist some hosts (google, hostmail, ...) because they are used from time to time to send spam.

I started looking into the policy_d-like features (https://github.com/vexim/vexim2/pull/153), perhaps that will be already efficient enough.

PR #163, opened 2016-07-18
add *_environment-fix

### LINKED ISSUE #140: Exim security fix CVE-2016-1531: environment
https://www.exim.org/static/doc/CVE-2016-1531.txt
It went into all debian releases and probably other distributions as well.
https://www.debian.org/security/2016/dsa-3517

Do we need any specific setting? I didn't go through all the postings on the mailinglist, some seem to run it without any additional settings.

Should we stick to the debian way and add it to the config-file?
````
.ifdef MAIN_KEEP_ENVIRONMENT
keep_environment = MAIN_KEEP_ENVIRONMENT
.else
keep_environment =
.endif
.ifdef MAIN_ADD_ENVIRONMENT
add_environment = MAIN_ADD_ENVIRONMENT
.endif
````

Udera:
fix #140

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -214,6 +214,14 @@ exim_user = mailnull
 exim_group = mail
 never_users = root
 
+.ifdef MAIN_KEEP_ENVIRONMENT
+keep_environment = MAIN_KEEP_ENVIRONMENT
+.else
+keep_environment =
+.endif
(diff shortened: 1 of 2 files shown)


Udera:
Thoughts about this? I have this running for several weeks now.

PR #170, opened 2016-07-24
Updates to the README.md for Debian

Udera:
I added some paths and specifics for debian. Furthermore I moved the mailman setup before exim because the install paths are needed and this way you don't have to go back and edit files twice.


README.md
```diff
@@ -103,9 +103,6 @@ FLUSH PRIVILEGES;
 ```
 Replace the word "CHANGE" with a secure password - It will be used both by the Exim MTA and the web interface to access the database.
 
-If the crypt() function on your system produces DES hashes, you will also have to uncomment the appropriate lines, as noted by the comments, at the end of mysql.sql.
-
-With the necessary changes made, you should run the following command line to initialize the database:
```
rimas-kudelis: Why remove this paragraph?

README.md
```diff
@@ -132,22 +129,22 @@ After copying the 'vexim' directory, you should find the 'variables.php.example'
 * $sqlpass – to the vexim database user's password which you chose while editing 'mysql.sql' in the "Databases and authentication" step.
 * $uid, $gid and $mailroot to the values you have from the "System user" step.
 * $cryptscheme is set to "sha512", a more specific configuration or other crypt-schemes can be used.
+* The URL to mailman in `$mailmanroot`
```
rimas-kudelis: why not write this the same way as other lines?
And if you want to add backticks, please add them to all of these variable names, not just this one.

README.md
```diff
@@ -205,6 +209,19 @@ The Diffie-Hellman group should have at least 1024 bit and can be created with t
...
 ```
+In the transport section, you need to modify the user and group to the exim-user `system_aliases`-transport (around line 800):
+```
+system_aliases:
+  driver = redirect
+  allow_fail
+  allow_defer
+  data = ${lookup{$local_part}lsearch{/etc/aliases}}
+  user = <exim-user>mailnull
+  group = <exim-group>mail
```
rimas-kudelis: what's this syntax?

README.md
Udera: I second was by accident. I didn't want to promote the clear-text version ...

README.md
```diff
@@ -101,11 +101,10 @@ GRANT SELECT,INSERT,DELETE,UPDATE ON `vexim`.* to "vexim"@"localhost"
     IDENTIFIED BY 'CHANGE';
 FLUSH PRIVILEGES;
 ```
-Replace the word "CHANGE" with a secure password - It will be used both by the Exim MTA and the web interface to access the database.
-
-If the crypt() function on your system produces DES hashes, you will also have to uncomment the appropriate lines, as noted by the comments, at the end of mysql.sql.
+Replace the word "CHANGE" with a secure password - It will be used both by the Exim MTA and the web interface to access the database. vexim uses SHA512-hashes for user passwords by default which is supported by current systems. You will find alternate hashing shemes (blowfish for *BSD-system or clear-text passwords) at the end of mysql.sql.
```
rimas-kudelis: Its Vexim, not vexim. Also, it should probably say "*BSD systems", and I think clear-text should probably not be mentioned at all.

Udera:
Ok. And I put all in one commit.

README.md
```diff
@@ -205,6 +211,19 @@ The Diffie-Hellman group should have at least 1024 bit and can be created with t
...
+```
+system_aliases:
+  driver = redirect
+  allow_fail
+  allow_defer
+  data = ${lookup{$local_part}lsearch{/etc/aliases}}
+  user = <exim-user>
+  group = <exim-group>
+  file_transport = address_file
+  pipe_transport = address_pipe
```
rimas-kudelis: 1. I think this is the router section, not transport. At least in Debian this is `conf.d/router/400_exim4-config_system_aliases`
2. Are these last four lines actually necessary? Again, in Debian they are all placed under `.ifdef`'s, hinting that they might be optional after all. If they are, we can just skip them, and then drop this section from the manual, right?

Udera: 1. Yes this is the router section.
2. Errors are often sent to system users. I like this feature a lot and mostly redirect everything to my mail address. For many services you could specify an address but forwarding is easier.

rimas-kudelis: 1. Then you should fix the first sentence.
2. Surely redirecting root mail via /etc/aliases is convenient, but how are these last four lines important for this? Like I said, this works fine in Debian, even though I don't have any of `SYSTEM_ALIASES_USER`, `SYSTEM_ALIASES_GROUP`, `SYSTEM_ALIASES_FILE_TRANSPORT`, `SYSTEM_ALIASES_PIPE_TRANSPORT` or `SYSTEM_ALIASES_DIRECTORY_TRANSPORT` defined.

Udera: 1. done
2. you won't need the last four lines for the redirection. But if you don't redirect, aren't these settings required to deliver mails to system users? If you haven't set a redirect yet, will it result in an error?

rimas-kudelis: I'm not sure I follow. This is a redirect router, so these lines apply to this particular router, don't they?
I'm not actually sure how this works, but it seems to work without these lines in Debian.

Udera: So you get mails delivered into a file of a local user? The debian config says:
````
# This router handles aliasing using a traditional /etc/aliases file.
# Delivery to arbitrary directories, files, and piping to programs in
# /etc/aliases is disabled per default.
````
All these conditional parts make the configuration file hard to read. You need to add to the documentation that if they want to use it, they have to define variables `SYSTEM_ALIAS_*` and where to put it.

Probably, most users only need the `/etc/alias` readout and we can comment the rest out. But this must be done in the configuration file.

Comment (Udera) [inline README.md]: > Probably, most users only need the /etc/alias readout and we can comment the rest out. But this must be done in the configuration file.

I will do this in a separate PR.

Comment (Udera) [inline README.md]: To improve the readability of the documentation, I'd like to mark debian, centOS or FreeBSD specific things. Not sure if we can do this by color or different boxes. So if you install on debian, you could ignore just all the other instructions.
Problem could be that we don't the precise instruction of each step in every system.

**Comment by rimas-kudelis** [comment]
This now need rebasing.

**Comment by Udera** [comment]
Yes, I need to rework over some more details ...

**Comment by Udera** [comment]
Ok, I rebased it, put in again the mailman-section and also removed the part about the username for system-users. I will comment this part in the configuration per default.

**Comment by rimas-kudelis** [comment]
I think we should use Debian as the basis for our manual. This way we wouldn't have to keep adding Debian-specific notes everywhere.

Or we could use RHEL/CentOS as the basis, in which case we could avoid adding the notes specific to that distro (or distro family).

Assuming that we aren't even sure if these default paths, groups and other specifics still apply to FreeBSD, upon which the manual was initially based, I don't think it makes sense to keep them in our default configuration.

File: README.md [hunk 6]
```diff
@@ -192,16 +207,13 @@ In order to add and delete domains from the database, you need to have a "site a
 
 The default username and password for the siteadmin, are:
 * **Username:** siteadmin
-* **Domain value:** (blank)
 * **Password:** CHANGE
```
rimas-kudelis: This needs updating as well now that #181 has been merged in.

README.md
```diff
@@ -192,16 +207,13 @@ In order to add and delete domains from the database, you need to have a "site a
...
 
-The password is case sensitive. You are strongly advised to log in and change it as soon as you get a chance. :-)
+The password is case sensitive. After you login, you have to change it first.
```
rimas-kudelis: I suggest to reword this as follows:
```
The password is case sensitive. You are advised to change it when you first log in.
```

Udera:
> I think we should use Debian as the basis for our manual. This way we wouldn't have to keep adding Debian-specific notes everywhere.
> 
> Or we could use RHEL/CentOS as the basis, in which case we could avoid adding the notes specific to that distro (or distro family).
> 
> Assuming that we aren't even sure if these default paths, groups and other specifics still apply to FreeBSD, upon which the manual was initially based, I don't think it makes sense to keep them in our default configuration.

That's a bit the ongoing discussion what we should do with our documentation. I played a bit with FreeBSD and the paths still turned out to be working (as far as I've seen), so I wouldn't delete this information.

README.md
```diff
@@ -74,15 +74,14 @@ You should create a new user account to whom the virtual mailboxes will belong.
...
 ```
-
 Where `YOUR_DATABASE_NAME` is the name of an empty database you have created for vexim. If you want the script to create the database for you and set up access to it, this is also doable: just open it in a text editor, and find a commented out block which begins with `-- CREATE DATABASE` near the top of the file. This block is documented just above it, so you may uncomment it, apply the changes you want and save the file. With the necessary changes made, you should run the following command line to initialize the database:
 ```
 # mysql -u root -p < vexim2/setup/mysql.sql
 ```
+A siteadmin-user is created with an autogenerated password, you will need it for your first login to the vexim-interface.
 
 ##### PGSQL:
 The code has been tested by several users to work with Virtual Exim, and we try our best to make sure it always will. Unfortunately I don't have much PostgreSQL knowledge to support it fully. A database schema for it is included however, as setup/pgsql.sql to help you set up the database. Make sure to adjust it similarly as per MySQL instructions above.
```
rimas-kudelis: Not necessarily now, but we should rewrite first person sentences (like `I don't have much foobar knowledge`) into more neutral ones.

Udera: Sure, a lot of parts need some improvement. I wouldn't change too much now if the instructions are clear and understandable and rather focus to complete the manual.

README.md
```diff
@@ -74,15 +74,14 @@ You should create a new user account to whom the virtual mailboxes will belong.
...
-
 ```
 # mysql -u root -D YOUR_DATABASE_NAME -p < vexim2/setup/mysql.sql
 ```
-
 Where `YOUR_DATABASE_NAME` is the name of an empty database you have created for vexim. If you want the script to create the database for you and set up access to it, this is also doable: just open it in a text editor, and find a commented out block which begins with `-- CREATE DATABASE` near the top of the file. This block is documented just above it, so you may uncomment it, apply the changes you want and save the file. With the necessary changes made, you should run the following command line to initialize the database:
 ```
 # mysql -u root -p < vexim2/setup/mysql.sql
 ```
+A siteadmin-user is created with an autogenerated password, you will need it for your first login to the vexim-interface.
```
rimas-kudelis: OK, I'll suggest some cosmetic changes.
siteadmin-user  → site administrator account OR site administrator user.
the vexim-interface → Vexim or the Vexim interface.

README.md
```diff
@@ -104,22 +103,32 @@ After copying the 'vexim' directory, you should find the 'variables.php.example'
 * $sqlpass – to the vexim database user's password which you chose while editing 'mysql.sql' in the "Databases and authentication" step.
 * $uid, $gid and $mailroot to the values you have from the "System user" step.
 * $cryptscheme is set to "sha512", a more specific configuration or other crypt-schemes can be used.
+* $mailmanroot to the default URL pattern of mailman
```
rimas-kudelis: $mailmanroot to the main Mailman URL.

README.md
```diff
@@ -128,6 +137,8 @@ Change this to the name of your server:
 ```
 primary_hostname=mail.example.org
 ```
+In general, it is required that your reverseDNS-entry of your IP points to this hostname.
```
rimas-kudelis: Is it really required?
either way:
reverseDNS-entry → reverse DNS entry

README.md
```diff
@@ -147,19 +158,24 @@ domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
 hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
 #trusted_users = www-data
 ```
-Specify here, the username and group under which Exim runs. This combination is also that under which mailman must run in order to work:
+This line configures database connectivity. You need to uncomment it and change the word 'CHANGE', to the password you will use for the 'vexim' database user, which we will set up in the next part. The socket path in Debian is `/var/run/mysqld/mysqld.sock` per default.
```
rimas-kudelis: This seems to be wrong. We've set up the user in Section one.

README.md
```diff
@@ -147,19 +158,24 @@ domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
...
 ```
-exim_user = mailnull
-exim_group = mail
+#hide mysql_servers = localhost::(/tmp/mysql.sock)/vexim/vexim/CHANGE
```
rimas-kudelis: Should paste the pgsql_servers line here as well.

I'm also thinking perhaps we should either change these lines to make it clear which part of them is the database name, and which is the user name, or add a short comment to the config file right above them.

README.md
```diff
@@ -147,19 +158,24 @@ domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
...
 ```
 # av_scanner = clamd:/tmp/clamd
 # spamd_address = 127.0.0.1 783
-``` 
-This line configures database connectivity. You need to uncomment it and change the word 'CHANGE', to the password you will use for the 'vexim' database user, which we will set up in the next part.
 ```
-#hide mysql_servers = localhost::(/tmp/mysql.sock)/vexim/vexim/CHANGE
+in Debian use:
+```
+# av_scanner = /var/run/clamav/clamd.ctl
```
rimas-kudelis: Shouldn't this be like this:
```
av_scanner = clamd:/var/run/clamav/clamd.ctl
```
?

README.md
```diff
@@ -190,18 +206,11 @@ Stefan Tomanek has a nice writeup about using Vexim in Debian, but that article
 #### Site Admin:
 In order to add and delete domains from the database, you need to have a "site admin". This user can create the initial postmaster users for the individual domains.
 
-The default username and password for the siteadmin, are:
-* **Username:** siteadmin
-* **Domain value:** (blank)
-* **Password:** CHANGE
-
-The password is case sensitive. You are strongly advised to log in and change it as soon as you get a chance. :-)
+The `siteadmin`-user has been created before with an auto-generated password (see mysql-section), use it here to login. The password is case sensitive. You are advised to change it when you first log in.
```
rimas-kudelis: This user has been created along with the database (refer to the Databases and authentication section above), use it here to log in. The password is case sensitive. You are advised to change it when you first log in.

And I think you can merge this paragraph with the previous one.

README.md
```diff
@@ -147,19 +158,25 @@ domainlist relay_to_domains = ${lookup mysql{RELAY_DOMAINS}}
 hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
 #trusted_users = www-data
 ```
-Specify here, the username and group under which Exim runs. This combination is also that under which mailman must run in order to work:
+These lines configure database connectivity. You need to uncomment one of them (depending on the database type you have chosen) and adjust it to match your setup. You have at least to change the word 'CHANGE' you used for the 'vexim' database user, which you have created before. The socket path depends on your system for Debian it is:
```
rimas-kudelis: Did you accidentally erase the words "to the password" ?
Also, please add a comma before "for Debian".

After this, I think you can merge. :)

Udera:
I moved outdated pop/imap configuration files into a subfolder `deprecated`

rimas-kudelis:
I think we can move Cyrus to the Wiki and kill others.

rimas-kudelis:
Here we go:
https://github.com/vexim/vexim2/wiki/Server-configuration:-Cyrus

Udera:
:+1:

rimas-kudelis:
Actually, if you copy your changes from the courier manual in source to Wiki as well, we can drop this whole clients directory.

Udera:
I was thinking doing the opposite. It's not a bad idea to keep the corresponding documentation with each release?

rimas-kudelis:
On the other hand, nevermind.

rimas-kudelis:
Exactly.
Another option would be to have a hierarchy of these manuals by release number (in the wiki), but we can figure this out later.

PR #173, opened 2016-07-30
Updated MySQL setup script

rimas-kudelis:
- All id columns have been changed to int(10) unsigned not null
- Foreign keys have been added to all tables referencing other tables
- Default `uid` and `gid` values in `domains` table have been changed to `65534`. This alleviates the need to edit this file manually before applying it. From looking at the code, it seems to me that `uid` and `gid` are always passed when creating a domain, so this requirement to edit the SQL default value is most likely unnecessary.

The SQL file itself was created using `mysqldump` and edited manually afterwards, that's why most of the file differs: `mysqldump` adds backticks around fields and indents them differently from what we had.

All FK's are currently ON UPDATE CASCADE ON DELETE CASCADE. This might need some tuning, although I'm not exactly sure.

@Udera: would you mind checking this out? I thought it might be nice to introduce these changes before finalizing the upgrade script to 2.3 in #72. On the other hand, these foreign key aren't that necessary and if we decide not to go that way in 2.3, I could make a much smaller alternative patch for point 3 only.

### DIFF
--- a/setup/mysql.sql
+++ b/setup/mysql.sql
@@ -1,160 +1,222 @@
+--
 -- Password for siteadmin is CHANGE. SHA512 is used by default, other
 -- crypting schemes can be found at the end of this file
 --
--- Database: `vexim`
---
-CREATE DATABASE IF NOT EXISTS `vexim` DEFAULT CHARACTER SET utf8;
+
+SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT;
+SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS;
+SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION;
+SET NAMES utf8;
+SET @OLD_TIME_ZONE=@@TIME_ZONE;
+SET TIME_ZONE='+00:00';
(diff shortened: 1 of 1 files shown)


Udera:
Great idea to remove the requirement to change the uid/gid in the SQL 
file.

I will try to get an old version of vexim installed on FreeBSD (it's my 
first time). They have ported a version which might already contain some 
fixes and then try the migration to vexim 2.3. And then a fresh setup of 
vexim 2.3.

rimas-kudelis:
So shall I merge this in right away?

Udera:
Why not. With too many PR open, it is difficult to test.

rimas-kudelis:
@Udera do you think it would be wrong to change the `vacation` field from `varchar(1024)` to `text`? I see no reason for limiting the size of that message, do you?

rimas-kudelis:
I added some more changes in 8cb216c5e73c6acaa3ac7d12152138e041b8fe10. Among them, dropped the commented out section which allowed to update siteadmin password to a bcrypt-hashed version. SHA512 is supported in FreeBSD since 10.0, and we now enforce password change on first login anyway, so it doesn't seem reasonable for me to bother the user with unnecessary instructions. Instead of that, I think it might be reasonable to have a note in the wiki explaining how to change the password manually via database if SHA512 is not supported for some reason.

Issue #174, opened 2016-08-01
Siteadmin password change requirement should be unavoidable


rimas-kudelis:
I just noticed that if you submit different passwords when required to change the siteadmin password, you are redirected to the domain list with a little message telling you the password has not been changed, and you aren't reminded about this until you log in the next time.

I think we should block any other actions while the password is unchanged instead.

rimas-kudelis:
This is now invalid via #181.

PR #178, opened 2016-08-02
Enforce siteadmin to change default password

Udera:
Fixes: https://github.com/vexim/vexim2/issues/174
Siteadmin must change his password and can't skip this step.

### DIFF
--- a/vexim/site.php
+++ b/vexim/site.php
@@ -12,6 +12,11 @@
   if (!isset($_POST['searchfor'])) {
     $_POST['searchfor'] = '';
   }
+  if(isset($_SESSION['pw_enforce'])) {
+    header ("Location: sitepassword.php");
+    die;
+  }
+
 ?>
 <html>
   <head>
--- a/vexim/sitepasswordsubmit.php
+++ b/vexim/sitepasswordsubmit.php
@@ -11,6 +11,7 @@
     $success = $sth->execute(array(':crypt'=>$cryptedpassword));
     if ($success) {
       $_SESSION['crypt'] = $cryptedpassword;
+      if(isset($_SESSION['pw_enforce'])) unset($_SESSION['pw_enforce']);
       header ("Location: site.php?sitepass=success");
       die;
     } else {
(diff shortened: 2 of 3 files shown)


vexim/sitepasswordsubmit.php
```diff
@@ -11,13 +11,18 @@
...
     } else {
       header ("Location: site.php?sitepass=fail");
       die;
     }
   } else {
+    if(isset($_SESSION['pw_enforce'])) {
+      header ("Location: sitepassword.php?badpass=siteadmin");
+      die;
+    }
     header ("Location: site.php?badpass=siteadmin");
```
rimas-kudelis: Perhaps we should just update this line instead?

vexim/site.php
```diff
@@ -12,6 +12,11 @@
   if (!isset($_POST['searchfor'])) {
     $_POST['searchfor'] = '';
   }
+  if(isset($_SESSION['pw_enforce'])) {
+    header ("Location: sitepassword.php");
+    die;
+  }
+
```
rimas-kudelis: I think we need similar precautions in `siteadd.php`, `siteaddsubmit.php`, `sitechange.php`, `sitechangesubmit.php` and `sitedelete.php`, don't we?

rimas-kudelis:
Hm, I have an alternative idea: why don't we use randomized password by default instead of hardcoding the word "CHANGE" ?

vexim/sitepasswordsubmit.php
Udera: Yes, could do it. However, if you changed your mind about changing the 
password (or it doesn't fit the security requirements of the 
password-policy), you are somehow trapped. You could just logout.

rimas-kudelis: If you change your mind, you can click on "Manage domains" and you're back to siteadmin's main view. If you however have entered two different passwords, or an insufficient one twice, you probably did this by mistake, and not on purpose just to get out of the "change password" window. So IMO returning the user back to the change password page is perhaps the sanest thing to do anyway.

Udera:
So should we remove all the checks for a `CHANGE` password?

rimas-kudelis:
> So should we remove all the checks for a `CHANGE` password?

Yes, I think so.

Udera:
Default password was removed (https://github.com/vexim/vexim2/pull/185). The initial checks have been removed, so closing this here.

PR #179, opened 2016-08-03
Update README.md

rimas-kudelis:
Update README.md to match the current status of mysql.sql

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This document provides a basic guide on how to get Virtual Exim working on your 
 
 Before we go into any details, I'd like to thanks Philip Hazel and the Exim developers for a fine product. I would also like to thanks the postmasters at various domains for letting me play havoc with their mail while I set this up :-) Finally, a special note of thanks to Dan Bernstein for his Qmail MTA. Dan, thank you for educating me how mail delivery really shouldn't be done, on the Internet.
 
-The Virtual Exim project currently lives on GitHub: https://github.com/avleen/vexim2
+The Virtual Exim project currently lives on GitHub: https://github.com/vexim/vexim2
 And its mailing list/Google group is available at: https://groups.google.com/group/vexim
 
 ### Installation steps for each component:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,47 +78,29 @@ You should create a new user account to whom the virtual mailboxes will belong.
...
-If the crypt() function on your system produces DES hashes, you will also have to uncomment the appropriate lines, as noted by the comments, at the end of mysql.sql.
-
-With the necessary changes made, you should run the following command line to initialize the database:
+Where `YOUR_DATABASE_NAME` is the name of an empty database you have created for vexim. If you want the script to create the database for you and set up access to it, this is also doable: just open it in a text editor, and find a commented out block which begins with `-- CREATE DATABASE` near the top of the file. This block is documented just above it, so you may uncomment it, apply the changes you want and save the file. With the necessary changes made, you should run the following command line to initialize the database:
 ```
 # mysql -u root -p < vexim2/setup/mysql.sql
 ```
 
+**NOTE:** if the `crypt()` function on your system does not support SHA-512 hashes, you will have to change the siteadmin password hash at the bottom of the file. Perhaps the easiest way is to set it to the following value: `$1$qZc7ANMc$h07fKA10jQQmJ33fzlJ3Z0`.
+
```
rimas-kudelis: @Udera, I'm thinking we should probably change the password hash supplied in `mysql.sql` back to MD5 scheme. This would allow us to drop this paragraph as well, by making the hash almost universally acceptable. I don't think anyone would go around changing the hashing scheme in the config from the default SHA-512 to MD5 just because they'd inspect the SQL file and find this string there. I tend to think if people won't know better, they'll just leave that option as is.

Udera: You mean the initial password is a MD5-hash which will then be changed 
to SHA512 when the siteadmin changes his password? So we only need to 
deal with encryption schemes in the configuration-file, which makes the 
installation process easier.

rimas-kudelis: Yes. Less gotchas → Shorter and easier-to-maintain manuals.

rimas-kudelis:
@Udera want to review, or shall I just merge this?

Udera:
Yes, it looks good. Makes the installation much easier 👍

PR #181, opened 2016-08-12
Use random password for siteadmin, output it nicely into terminal

rimas-kudelis:
This is an alternative for #178. With random passwords in place, checks for `$password === 'CHANGE'` would become obsolete.

@Udera, what do you think?

### DIFF
--- a/setup/mysql.sql
+++ b/setup/mysql.sql
@@ -1,3 +1,10 @@
+-- The following defines the salt prefix for siteadmin password hash..
+-- When adding the siteadmin user, we will hash that user's password using a secure hashing scheme
+-- (SHA-512 by default, supported by Linux, FreeBSD and Solaris).
+-- Below you may change this prefix if you prefer a different hashing scheme, e.g. '$2a$10$' for bcrypt
+-- (*BSD-only Blowfish hashing scheme):
+SET @PW_PREFIX='$6$';
+
 --
 -- The password for siteadmin is CHANGE.
 --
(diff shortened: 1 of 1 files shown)


setup/mysql.sql
```diff
@@ -198,12 +205,18 @@ ALTER TABLE `domains` ENABLE KEYS;
...
+SET @SITE_ADMIN_PW = SUBSTRING(TO_BASE64(LOWER(CONV(FLOOR(RAND() * 99999999999999), 10, 36))),1,10);
+
+--
 -- Seed the `users` table with the siteadmin user
 --
 
 LOCK TABLES `users` WRITE;
 ALTER TABLE `users` DISABLE KEYS;
-INSERT INTO `users` VALUES (1,1,'siteadmin','siteadmin','$6$uR.EiB1dj5rrvwMF$Qh5LgdjOZavKXwhi9IF0Yuzu7qxsG.dLTTB8e./55ZRNfBuZVLnfUSOEXa0oWT6174myO.WYkOy83HYWAKPbK/',65535,65535,'','','site',1,0,0,0,0,0,0,0,1,NULL,NULL,0,0,0,'SiteAdmin',0,0,NULL,NULL);
+INSERT INTO `users` VALUES (1,1,'siteadmin','siteadmin',ENCRYPT(@SITE_ADMIN_PW,CONCAT(@PW_PREFIX,MD5(RAND()))),65535,65535,'','','site',1,0,0,0,0,0,0,0,1,NULL,NULL,0,0,0,'SiteAdmin',0,0,NULL,NULL);
```
rimas-kudelis: There are some issues with this:

1. I've again defined the password hash prefix at the top of the file, so we would probably have to keep the paragraph about changing it if necessary (or we could use an MD5 hash instead). If we went with the MD5 hash, we should probably add a call in Vexim to update the hash upon login if its prefix does not match what's in the settings. On the other hand, we don't *have* to fix this, I think this is quite secure already, and we encourage the user to change the password anyway.
2. The salt is not as long as it could be with SHA-512. To fix this, we could `CONCAT()` the results of several `MD5(RAND())` calls. I would also probably enclose these in a `TO_BASE64()` call, to have mixed-case salt.
3. This generator doesn't add any special characters into the password or the salt. It is possible to add these as well, but I think defining a huge MySQL function for just this one password is probably overkill.
4. `ENCRYPT()` is deprecated. Although this does not have immediate effect, we should not forget this.

Udera: > I think this is quite secure already, and we encourage the user to change the password anyway.

There is a risk that the user will keep this password because it "looks" secure.



> If we went with the MD5 hash, we should probably add a call in Vexim to update the hash upon login if its prefix does not match what's in the settings.

If we add this to vexim, also user passwords would be updated, perhaps not a bad idea? On the other side, a sha512-pw with a longer hash does require only some work on the mysql-creation-scripts, so less testing.



> ENCRYPT() is deprecated. Although this does not have immediate effect, we should not forget this.

For vexim3, I'd like to see some kind of installer-script in php. So it creates the database for you.

rimas-kudelis: > There is a risk that the user will keep this password because it "looks" secure.

OK, let's leave the prefix as-is. After all, *most* systems support SHA-512.

>On the other side, a sha512-pw with a longer hash does require only some work on the mysql-creation-scripts, so less testing.

Actually, I just checked and the salt is long enough, or at least it seems to be in most cases (above 40 characters, whereas even `bcrypt` needs just 22, and `SHA-512` [needs 16](http://php.net/crypt)).

I've just updated the PR to have mixed-case salts. I think this is good to go.

> For vexim3, I'd like to see some kind of installer-script in php. So it creates the database for you.

Me too.

Udera:
Good idea. What about postgres, do they support that as well?

rimas-kudelis:
No idea, but I could investigate.
Or we could say we only support MySQL for now. :) Lame, I know. :)

Udera:
Tested on freebsd 👍

rimas-kudelis:
> What about postgres, do they support that as well?

They seem to support it after enabling the `pgcrypto` extension for the database. This action requires superuser privileges. Not sure if it's worth it, considering that we only need this functionality once.

rimas-kudelis:
On the other hand, if we keep the `CREATE DATABASE` line in PostgreSQL file, it will require superuser privileges anyway, in which case It's probably OK to enable that extension for the DB.
I'll try this approach, and see how it goes.

PR #182, opened 2016-08-14
Set default values if spamassassin is disabled

### LINKED ISSUE #180: adding forwarding address(es)
Hi Guys,

I have a new server running FreeBSD 10.2 and exim 4.8. I have vexim2 installed and all seems to be working fine. However I have noticed that when you try to add a forwarding address on a user/address, you can only add one address contrary to the previous vexim where you could add multiple addresses. 
I get this error : "user1@xyz.com,user2@abc.com is not a valid e-mail address" when i separate with a comma. Have tried using a semi colon with the same error.

Udera:
This was no error of the forwarded addresses, the problem was, if spamassassin is disabled, the form does not transmit a `$_POST['sa_tag']` (also `$_POST['sa_refuse']` and `$_POST['spam_drop']$_POST['spam_drop']`) value which then is `NULL`. But the database does not allow `NULL`. In `adminuseraddsubmit.php` we already fixed that problem but it slipped for alias-addresses.

I improved the initial approach, which set the `sa_tag` to `0`. I use instead the `$sa_tag` from the `variables.php`.

@mistatongz: can you test if that solves your problem?


rimas-kudelis:
I find it hard to believe that this fixes #180, which seems like a totally different problem. Other than that, I think the patch is fine.

Udera:
Yes, probably a coincidence. I didn't get the same error. The creation of a mail account just failed. Now I can add multiple addresses to forward (on FreeBSD test machine).

PR #183, opened 2016-08-14
Replace dnsbl.org with ix.dnsbl.manitu.net

rimas-kudelis:
This also disables DNSBL checks for authenticated users.
This is a copy of #127, but has a reworded comment block about disabled DNSBL checks. I think this wording describes the situation better.
Fixes #127


rimas-kudelis:
Part of this is already fixed in #166.

PR #187, opened 2016-08-14
Synchronized PostgreSQL setup script with the MySQL script

rimas-kudelis:
This updates the PostgreSQL setup script to match the MySQL one. I've tested and it works, or at least is successfully being executed with `psql`.

I haven't added the random siteadmin password feature though: the `CRYPT()` function in PostgreSQL [supports only a small number of hashing methods](https://www.postgresql.org/docs/9.5/static/pgcrypto.html#AEN172965), of which only MD5 is universally supported. I tried passing an SHA-512 salt to `CRYPT()`, but it falls back to the DES algorithm in that case. Bcrypt is supported, but it's not supported by Linux, so it's not exactly an option.

I figured out it's better to set the password to our traditionally used 'CHANGE' in this case, with hopes that the siteadmin will change it.

### DIFF
--- a/setup/pgsql.sql
+++ b/setup/pgsql.sql
@@ -1,86 +1,278 @@
-CREATE DATABASE vexim WITH ENCODING 'UTF8';
-CREATE TABLE domains (domain_id SERIAL PRIMARY KEY,
-	domain varchar(255) UNIQUE NOT NULL,
-	maildir varchar(4096) NOT NULL default '',
-	uid int NOT NULL default '65534' CHECK(uid BETWEEN 1 AND 65535),
-	gid int NOT NULL default '65534' CHECK(uid BETWEEN 1 AND 65535),
-	max_accounts int NOT NULL default '0',
-	type varchar(5) default NULL,
-	avscan smallint NOT NULL default '0',
-	blocklists smallint NOT NULL default '0',
-	enabled smallint NOT NULL default '1',
-	mailinglists smallint NOT NULL default '0',
-	pipe smallint NOT NULL default '0',
-	spamassassin smallint NOT NULL default '0',
-	quotas int NOT NULL default '0' CHECK(quotas > -1),
-	maxmsgsize int NOT NULL default '0' CHECK(maxmsgsize > -1),
-	sa_tag int NOT NULL default '0' CHECK(sa_tag > -1),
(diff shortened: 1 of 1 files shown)


Udera:
Or we use `md5()` to get a similar installation routine? The admin should change the password as well... what is less likely as for CHANGE. Not sure about it, I was so happy that you managed to get rid of 'CHANGE' 😄

rimas-kudelis:
I liked it too. I just thought it would be more likely that the admin would change the password which is CHANGE than the random password. :)
I think it's possible to use the MD5 hash like you suggest, with an option to switch to a bcrypt-based hash, and with a suggestion to change the password on first login anyway.

rimas-kudelis:
(closed accidentally)

rimas-kudelis:
I'll make an alternative PR if I succeed in preparing the script with random passwords. We can choose then.

rimas-kudelis:
After opening another PR, I decided to actually update this one instead.
So here we have the PostgreSQL setup script which generates random password for siteadmin if pgcrypto extension is enabled (or can be enabled).

Note that since pgcrypto is not guaranteed to be enabled, and the script has database and user creation commented out (thus is designed to be run by a non-elevated target user by default), it's possible that we won't be able to encrypt the password. In that case, our old default 'CHANGE' will be used.

I'm still unsure if it's better to ship the scripts with DB and user creation commented out or not. @Udera, what do you think?

I'll play a bit more to see if I can make MySQL and PostgreSQL use random passwords when creating database users as well, but this can be done in a separate PR. So far, I tried a little and haven't succeeded.

Udera:
In general, I think it's a good idea to create the db-user as well and set the right permissions. I don't know if you can create random-pw with pure mysql/PostgreSQL-magic. It is certainly the way to go for the future version with an installer script, if you find a way here, perfect. Or can you request some user input from command line when executing the script?

Udera:
To keep the user-creation commented would be fine as well. Advantage of it being commented: Users must actively open the file, uncomment it and then hopefully change the default password.

I'd prefer to release 2.3.0 soon and move some minor improvements in a 2.3.1 version.

rimas-kudelis:
User/DB creation is currently commented out in the MySQL file, and I think we should keep its status in sync between the two files, so let's merge this as-is as well.

Udera:
Allright.

PR #188, opened 2016-08-14
Update db schemas

rimas-kudelis:
Some late changes to the `users` table:

- column `forward`: changed from `VARCHAR(255)` to `VARCHAR(4096)` (as requested by @runout-at)
- column `unseen`: added a `NOT NULL` constraint
- column `vacation`: changed from `VARCHAR(1024)` to `TEXT`

This should be applied after #187, the diff will be much smaller then. :)


Udera:
If we restrict the column `unseen`, we should check the scripts if they handle this restriction correctly (adminuser..., userchange...).

rimas-kudelis:
Isn't it processed in the same way as other columns?

rimas-kudelis:
I just checked my setup, and there are no rows matching the `unseen IS NULL` criteria.
I also grepped our sources for this word, and wherever we touch this field, we always set it to either 1 or 0, never NULL. I don't think this change to `unseen` is dangerous.

Udera:
The only dangerous fields are optional fields that can be disabled for a domain (sa, antivirus, pipe, ...). `unseen` is always an option, so it shouldn't matter.

Udera:
This branch has conflicts, can you rebase?

rimas-kudelis:
Can we resolve #187 first?

rimas-kudelis:
Or do you want to merge in this one first, and then go to #187? Either way suits me, I just have to know what to rebase on what. :)

Udera:
Let's do #187 first.

rimas-kudelis:
PR updated.

PR #189, opened 2016-08-15
Synchronized PostgreSQL setup script with the MySQL script

rimas-kudelis:
This also makes PostgreSQL setup generate random password for siteadmin if pgcrypto extension is enabled (or can be enabled).

This is an alternative to #187, this time with the random password possibility.
Note that since pgcrypto is not guaranteed to be enabled, and the script has database and user creation commented out, it's possible that we won't be able to encrypt the password. In that case, our old default 'CHANGE' will be used.

I'm still unsure if it's better to ship the scripts with DB and user creation commented out or not. @Udera, what do you think? If we ship them with these lines uncommented, I guess I could improve them to use a randomized password for the database user as well, and output that password along with siteadmin's.

### DIFF
--- a/setup/pgsql.sql
+++ b/setup/pgsql.sql
@@ -1,86 +1,278 @@
-CREATE DATABASE vexim WITH ENCODING 'UTF8';
-CREATE TABLE domains (domain_id SERIAL PRIMARY KEY,
-	domain varchar(255) UNIQUE NOT NULL,
-	maildir varchar(4096) NOT NULL default '',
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
> If we ship them with these lines uncommented, I guess I could improve them to use a randomized password for the database user as well, and output that password along with siteadmin's.

Meh, this seems more difficult than I expected, not sure if possible. I'm not doing this.

rimas-kudelis:
Let's continue discussion on #187 instead.

PR #202, opened 2016-08-18
Replace `imap_qprint()` and `imap_8bit()` with `quoted_printable_[de|en]code()`

rimas-kudelis:
Instead of using these functions provided by the IMAP module, which might be unavailable, we can use `quoted_printable_encode()` and `quoted_printable_decode()`. This means that we'll only depend on the IMAP module to check user's quota from now on.

### DIFF
--- a/vexim/adminuserchange.php
+++ b/vexim/adminuserchange.php
@@ -268,17 +268,10 @@
           </td>
         </tr>
         <tr>
-        <?php if (function_exists('imap_qprint')) { ?>
           <td><?php echo _('Vacation message'); ?>:</td>
           <td>
-            <textarea name="vacation" cols="40" rows="5" class="textfield"><?php print imap_qprint($row['vacation']); ?></textarea>
+            <textarea name="vacation" cols="40" rows="5" class="textfield"><?php print quoted_printable_decode($row['vacation']); ?></textarea>
           </td>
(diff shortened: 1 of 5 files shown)


Udera:
If I use this vacation message:
````
ä
ö
ü
ß
````
I get the correct reply with the current version. With your modification, I get:
````
=E4
=F6
=FC
=DF
````
In vexim it is always shown correct. There is some encoding issue, but I couldn't figure out where and why.

**Comment by rimas-kudelis** [comment]
It's weird that you get escape codes for single-byte characters instead of double-byte.
Couple questions:

1. are you testing with latest vexim? We have the following line in config/httpheaders.php, so your vacation should be stored in QP-encoded UTF-8, not QP-encoded ISO-8859-1 in the database:
```
header("Content-Type: text/html; charset=utf-8");
```
2.  Do you have the following line in your vacation transport:
```
headers  = "Content-Type: text/plain; charset=UTF-8\nContent-Transfer-Encoding: quoted-printable"
``` ? If you do, your e-mail client should be able to undersand that the vacation message needs qp-decoding.

Udera:
I used the latest master and your pull request. I didn't update the whole exim4.conf, most was an old part without the encoding-headers. And php-imap was missing as well. Now everything works as it is supposed to.

PR #203, opened 2016-08-19
Do not add spam headers during ACL phase so we don't have to remove them later

rimas-kudelis:
Instead, persist `$spam_*` variables as `$acl_m_spam_*` (except `$spam_score_int`, which persists by itself), and use these variables to add headers at delivery time.
This will add `X-Spam-Score`, `X-Spam-Score-Int` and `X-Spam-Bar` headers regardless of whether or not the message is spam, but this can be easily changed.

Comments welcome.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -87,9 +87,13 @@ hostlist   relay_from_hosts = localhost : @ : 192.168.0.0/24
 # Comment this line if you want to disable it, instead of + you can use a different separator.
 VEXIM_LOCALPART_SUFFIX = +*
 
-# if you exim version < 4.83 , define macro below for enabling new
-# 'headers_remove' behavior. See https://github.com/vexim/vexim2/pull/102 .
-#OLD_HEADERS_REMOVE = yes
+# Exim will put the detailed spam report into an X-Spam-Report header by default.
+# This report is really huge by default, but its template can be tweaked to make
+# it look almost exactly like contents of the X-Spam-Status header, which
+# SpamAssassin adds when scanning messages externally, and which is a much more
+# compact version of the report. If you tweak your template this way, you may
+# as well want to change the header name here.
+#VEXIM_SPAM_REPORT_HEADER_NAME = X-Spam-Status
 
 # Reverse DNS checks can be efficient to fight spam. It checks if host name associated to the ip address matches
 # the sender_host_name (for properly configured mail servers this should be the case). However, there are some
@@ -463,6 +467,11 @@ acl_check_content:
 
 begin routers
 
+# Make sure that the name of the spam report header is set.
+.ifndef VEXIM_SPAM_REPORT_HEADER_NAME
+VEXIM_SPAM_REPORT_HEADER_NAME = X-Spam-Report
+.endif
+
 # This router routes to remote hosts over SMTP by explicit IP address,
 # when an email address is given in "domain literal" form, for example,
 # <user@[192.168.35.64]>. The RFCs require this facility. However, it is
@@ -476,7 +485,6 @@ begin routers
 #   driver = ipliteral
 #   domains = ! +local_domains
 #   transport = remote_smtp
-#   headers_remove = X-Spam-Report:X-Spam-Score
 
 
 # This router routes addresses that are not in local domains by doing a DNS
@@ -492,7 +500,6 @@ dnslookup:
   domains = ! +local_domains
   transport = remote_smtp
   ignore_target_hosts = 0.0.0.0 : 127.0.0.0/8
-  headers_remove = X-Spam-Report:X-Spam-Score
   no_more
 
 mailman_router:
@@ -502,7 +509,6 @@ mailman_router:
   local_part_suffix = -bounces : -bounces+* : \
                       -confirm+* : -join : -leave : \
                       -owner : -request : -admin
-  headers_remove = X-Spam-Score:X-Spam-Report
   transport = mailman_transport
 
 # The remaining routers handle addresses in the local domain(s).
@@ -597,6 +603,10 @@ ditch_spam:
         and users.domain_id=domains.domain_id \ 
         and users.sa_refuse > 0 }{$value}fail}} {yes}{no}}
     retry_use_local_part
+    headers_add = ${if and { \
+                     {match{$domain}{$original_domain}} \
+                     {match{$local_part}{$original_local_part}} \
+                     } {X-Spam-Flag: YES\nX-Spam-Score: $acl_m_spam_score\nVEXIM_SPAM_REPORT_HEADER_NAME: $acl_m_spam_report}{} }
 
 ditch_hdrmailer:
   driver = redirect
@@ -715,31 +725,11 @@ virtual_domains:
   allow_fail
   data = ${extract{smtp}{$address_data}}
   headers_add = ${if and { \
+                    {match{$domain}{$original_domain}} \
+                    {match{$local_part}{$original_local_part}} \
                     {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
                     {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
-                    } {X-Spam-Flag: YES\n}{} }
-  # Whether to use old or new headers_remove behavior.
-  .ifndef OLD_HEADERS_REMOVE
-    headers_remove = ${if or { { <{$spam_score_int}{1} } \
-                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
-                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
-                                    } \
-                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
-                            }  {X-Spam-Score}}
-    headers_remove = ${if or { { <{$spam_score_int}{1} } \
-                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
-                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
-                                    } \
-                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
-                            }  {X-Spam-Report}}
-  .else
-    headers_remove = ${if or { { <{$spam_score_int}{1} } \
-                              { and {{<{$spam_score_int}{${extract{sa_tag}{$address_data}}} } \
-                                     {eq {1}{${extract{on_spamassassin}{$address_data}}}} \
-                                    } \
-                              } { eq {0}{${extract{on_spamassassin}{$address_data}}}} \
-                            }  {X-Spam-Score:X-Spam-Report}}
-  .endif
+                    } {X-Spam-Flag: YES\nX-Spam-Score: $acl_m_spam_score\nVEXIM_SPAM_REPORT_HEADER_NAME: $acl_m_spam_report}{} }
   # using local_part_suffixes enables possibility to use user-"something" localparts
   # which could cause you trouble if you're creating email-adresses with dashes in between.
   .ifdef VEXIM_LOCALPART_SUFFIX
--- a/docs/configure
+++ b/docs/configure
@@ -455,7 +455,7 @@ acl_check_content:
 
   # finally accept all the rest
   accept
-  
+
 
 ######################################################################
 #                      ROUTERS CONFIGURATION                         #
@@ -560,7 +560,7 @@ ditch_malware:
   condition = ${if and { {match {$h_X-ACL-Warn:}{.*malware.*}} \
                          {eq {${lookup mysql{select users.on_avscan from users,domains \
 		                where localpart = '${quote_mysql:$local_part}' \
-		                and domain = '${quote_mysql:$domain}' \ 
+		                and domain = '${quote_mysql:$domain}' \
 		                and users.on_avscan = '1' \
                                 and domains.avscan = '1' \
 		                and users.domain_id=domains.domain_id}}}{1} }} {yes}{no} }
@@ -584,23 +584,23 @@ ditch_spam:
     driver = redirect
     allow_fail
     file_transport = virtual_ditch_spam_transport
-    data = ${lookup mysql{select concat(smtp,'/.Spam') \ 
-        from users,domains \ 
-        where localpart = '${quote_mysql:$local_part}' \ 
-        and domain = '${quote_mysql:$domain}' \ 
-        and domains.enabled = '1' \ 
-        and users.enabled = '1' \ 
+    data = ${lookup mysql{select concat(smtp,'/.Spam') \
+        from users,domains \
+        where localpart = '${quote_mysql:$local_part}' \
+        and domain = '${quote_mysql:$domain}' \
+        and domains.enabled = '1' \
+        and users.enabled = '1' \
         and users.domain_id = domains.domain_id}}
-    condition = ${if >{$spam_score_int}{${lookup mysql{select \ 
-        users.sa_refuse * 10 from users,domains \ 
-        where localpart = '${quote_mysql:$local_part}' \ 
-        and domain = '${quote_mysql:$domain}' \ 
-        and users.on_spamassassin = '1' \ 
+    condition = ${if >{$spam_score_int}{${lookup mysql{select \
+        users.sa_refuse * 10 from users,domains \
+        where localpart = '${quote_mysql:$local_part}' \
+        and domain = '${quote_mysql:$domain}' \
+        and users.on_spamassassin = '1' \
         and domains.spamassassin = '1' \
         and users.spam_drop = '0' \
         and users.on_forward = '0' \
         and users.type = 'local' \
-        and users.domain_id=domains.domain_id \ 
+        and users.domain_id=domains.domain_id \
         and users.sa_refuse > 0 }{$value}fail}} {yes}{no}}
     retry_use_local_part
     headers_add = ${if and { \
@@ -681,7 +681,7 @@ virtual_vacation:
   condition = ${if and { {!match {$h_precedence:}{(?i)junk|bulk|list}} \
                          {eq {${lookup mysql{select users.on_vacation from users,domains \
 		                where localpart = '${quote_mysql:$local_part}' \
-		                and domain = '${quote_mysql:$domain}' \ 
+		                and domain = '${quote_mysql:$domain}' \
 		                and users.on_vacation = '1' \
 		                and users.domain_id=domains.domain_id}}}{1} }} {yes}{no} }
   no_verify
@@ -706,7 +706,7 @@ virtual_forward:
   condition = ${if and { {!match {$h_precedence:}{(?i)junk}} \
                          {eq {${lookup mysql{select users.on_forward from users,domains \
 		                where localpart = '${quote_mysql:$local_part}' \
-		                and domain = '${quote_mysql:$domain}' \ 
+		                and domain = '${quote_mysql:$domain}' \
 		                and users.on_forward = '1' \
 		                and users.domain_id=domains.domain_id}}}{1} }} {yes}{no} }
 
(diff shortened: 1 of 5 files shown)


rimas-kudelis:
I did some whitespace shuffling in `vexim-acl-check-content.conf`, so you may want to use ?w=0 when looking at the diff. Sorry. I can split this into two commits, but these whitespace changes aren't that big after all, so I was like "meh, nevermind". :)

docs/configure
```diff
@@ -714,32 +707,13 @@ virtual_domains:
                 and users.domain_id = domains.domain_id}{$value}fail}
   allow_fail
   data = ${extract{smtp}{$address_data}}
+  headers_add = X-Spam-Score: $acl_m_spam_score
+  headers_add = X-Spam-Score-Int: $spam_score_int
+  headers_add = X-Spam-Bar: $acl_m_spam_bar
```
Udera: If you forward e-mail to your account, these headers will be added twice. Is it possible to put in a condition to add the header only if there isn't such a header already?

Udera: And shouldn't they only be added if spam score is larger than `sa_tag`?

rimas-kudelis:
> If you forward e-mail to your account, these headers will be added twice. Is it possible to put in a condition to add the header only if there isn't such a header already?

I didn't think about it. I think we could unconditionally remove headers before adding them.

> And shouldn't they only be added if spam score is larger than `sa_tag`?

I mentioned this in my initial comment. :)

docs/configure
rimas-kudelis: > If you forward e-mail to your account, these headers will be added twice. Is it possible to put in a condition to add the header only if there isn't such a header already?

I didn't think about it. I think we could unconditionally remove headers before adding them.

> And shouldn't they only be added if spam score is larger than `sa_tag`?

I mentioned this in my initial comment. :)

rimas-kudelis: Actually, what do you mean by "If you forward e-mail to your account"?

docs/configure
```diff
@@ -492,7 +487,6 @@ dnslookup:
   domains = ! +local_domains
   transport = remote_smtp
   ignore_target_hosts = 0.0.0.0 : 127.0.0.0/8
-  headers_remove = X-Spam-Report:X-Spam-Score
```
Udera: Not sure if it is a great idea to keep the `X-Spam-*`-headers in. Or we should modify them `X-Spam-*-MYDOMAIN`.

Udera:
And you forgot to add spam-headers in the router `ditch_spam` where this is of special interest to debug false positives.

docs/configure
Udera: You have set up a forwarding address: forward@example.org -> account@example.org. For both addresses, you have enable spamassassin.

Other idea: Only add X-Spam-Flag and X-Spam-Report on real mail accounts (and not on forwarded addresses at all). Forwarded addresses are either delivered to a local mailbox (and then this setting should decide what happens), or they are delivered to an external mailserver, there we do not want this tag either (or we need to modify it).

docs/configure
rimas-kudelis: Note that we haven't added these headers yet, so if they exist, they were sent to us. I agree it doesn't make sense for the sender to send them, but does it make sense for us to remove them?

rimas-kudelis:
Indeed, added them now.

docs/configure
Udera: For forwarding addresses? forward@example.org -> myaddress@gmail.com

Udera: I'm just wondering what is best practise for spam-headers. Imagine you get external e-mails with `X-Spam-Flag: YES`. Now you pass mails directly via pipe to procmail or dovecot who do the filtering for you. To avoid confusion with tags our system adds, you must either remove or rename such tags. On the other hand, if the email was already considered as spam by the delivering mail server why accepting it in the first place?

rimas-kudelis: I'm actually doing this right now (passing all messages, regardless of spam status), except I'm not passing on these headers.

rimas-kudelis: Should I add a `remove_header` modifier right in the ACL? Or should I add modifiers in the routers instead?

rimas-kudelis: Would this in `vexim_acl_check_content.conf` be sufficient:
```
deny message = "Sending server openly considers message as spam ('X-Spam-Flag: YES' found)."
regex = ^X-Spam-Flag: *YES
```
?

rimas-kudelis:
PR updated

rimas-kudelis:
Huh, I just realized this is 99% same as @Z3po's proposal in #102... :)

docs/debian-conf.d/router/250_vexim_virtual_domains
```diff
@@ -52,32 +52,11 @@ virtual_domains:
                 and users.domain_id = domains.domain_id}{$value}fail}
   allow_fail
   data = ${extract{smtp}{$address_data}}
+  headers_add = ${if eq{$spam_score_int}{} {}{X-Spam-Score: $acl_m_spam_score}}
```
rimas-kudelis: Also might be worth considering:
```
  headers_add = ${if eq{$spam_score_int}{} {X-Spam-Checked: no}{X-Spam-Score: $acl_m_spam_score}}
```

docs/vexim-acl-check-content.conf
```diff
@@ -1,44 +1,59 @@
...
-        log_message	= This message contains malware ($malware_name)
+  #
+  warn    message     = This message contains malware ($malware_name)
+          malware     = *
+          log_message = This message contains malware ($malware_name)
+
+  # Reject messages with an "X-Spam-Flag: YES" header.
+  #
+  deny    message = Sending server openly considers message as spam ('$regex_match_string' found).
+          regex   = ^X-Spam-Flag:: *YES
```
rimas-kudelis: There's one problem with this: it might match body as well. But I don't think this is a big issue.

docs/configure
Udera: I found more about the headers (https://github.com/Exim/exim/wiki/AclSmtpData), so we could use:
````
condition       = ${if def:h_x-spam-flag: {yes}{no}}
````
this only checks if the header is present, then we would need a second condition:
````
${if bool{$h_x-spam-flag}}
````
perhaps it's just enough to use:
````
${if bool_lax{$h_x-spam-flag}}
````
http://www.exim.org/exim-html-current/doc/html/spec_html/ch-string_expansions.html

Comment (rimas-kudelis) [inline docs/configure]: I think `bool` alone might be enough.

Comment (rimas-kudelis) [inline docs/configure]: I've amended the commit

Comment (Udera) [inline docs/configure]: I haven't setup spamassassin on my latest FreeBSD-test system yet. I'd 
like to test this before I use it on a productive system.

Comment (rimas-kudelis) [inline docs/configure]: go on then :)

I think I'd like to also explain how to compact `X-Spam-Report` to look like `X-Spam-Status`  somewhere in the config/manual, cuz I really love that bit. At the very least message preview is totally useless in that header.

BTW should we remove other headers usually added by SpamAssassin from received messages as well?

Comment (Udera) [inline docs/configure]: Do you have a larger mailbox? You could just grep through the mails, if 
there are many with spamassassin-stuff from other mail hosts. I will try 
that as well.

Comment (rimas-kudelis) [inline docs/configure]: > Do you have a larger mailbox? 

Yes

> You could just grep through the mails, if there are many with spamassassin-stuff from other mail hosts. I will try that as well.

I have messages with these headers dating back to 2006, but I can't know for sure whether or not it was my server that added them.

I'm just thinking if we're removing these headers, why not remove more.

On the other hand, perhaps we shouldn't be removing them after all, since preserving them might help with debugging etc. The important thing is not to consider them authoritative. I just got an e-mail from LibreOffice mailing list, and it contains all these X-Spam headers from LibO server.

The only downside of keeping them I can think of right now is that they may make the decision more difficult for MUA's which can be told to treat them as authoritative. If one `X-Spam-Flag` says "NO", and the other says "YES", which one is the right one?
On the other hand, mail server admin might be about the only person with a "my server knows best" opinion, everyone else might be more open or not care enough.

I'm now in favor of at least commenting out removal of these headers by default. :)

Comment (Udera) [inline docs/configure]: For this reason, I currently use this header to set the filter:
`X-Spam-Report: Spam detection software, running on the system "example.org",`
the `X-Spam-Report`-tag in combination with `example.org` is unique for my mail server. I am not happy with that at all.

Why don't I use the `X-Spam-Flag`? So far, I only filter mail above the sa_refuse-score (like the `ditch_spam`-router only moves mails above this limit in the spam folder). For training purposes, I have set up my mail account:
sa > sa_refuse -> `Spam`
sa_refuse > sa > sa_tag -> `suspicious`
sa < sa_tag -> `inbox`

Not sure, how we should handle this for all users. In my case, everything above sa_refuse is nearly 100% spam. From time to time, there are some newsletters, registration mails in my `suspicious`-folder, but >90% is spam. But two spam-folders are too much for most users. If everybody uses the same `sa_refuse`-score, I can refuse this mails straight away.

Comment (rimas-kudelis) [inline docs/configure]: > For this reason, I currently use this header to set the filter:
> `X-Spam-Report: Spam detection software, running on the system "example.org",`
> the `X-Spam-Report`-tag in combination with `example.org` is unique for my mail server. I am not happy with that at all.

Why's this bad?
Regarding other things, I don't think I have enough knowledge to talk. After I migrated my setup to 2.3 couple days ago, my spam filtering became worse. And I don't even know how to start fixing that...

Comment (Udera) [inline docs/configure]: >> For this reason, I currently use this header to set the filter:
> X-Spam-Report: Spam detection software, running on the system "example.org",
> the X-Spam-Report-tag in combination with example.org is unique for my mail server. I am not happy with that at all.

> Why's this bad?

It's working, however you have to be careful about this rules. Moving the server, I had to change the host name and such settings are very likely to be missed.




> After I migrated my setup to 2.3 couple days ago, my spam filtering became worse. And I don't even know how to start fixing that...

That is bad. You wrote in a different post, that you run sa-exim before, you are sure that you run spamassassin as the same user with the same settings? Or you had more ACL-checks in use which are not part of the current vexim setup (SPF, rDNS checks, ...)?

Comment (rimas-kudelis) [inline docs/configure]: > It's working, however you have to be careful about this rules. Moving the server, I had to change the host name and such settings are very likely to be missed.

Migrations are always tough. :)

> That is bad. You wrote in a different post, that you run sa-exim before, you are sure that you run spamassassin as the same user with the same settings? Or you had more ACL-checks in use which are not part of the current vexim setup (SPF, rDNS checks, ...)?

Here's what I found in sa-exim config file:
```
# SAspamcUser: The username passed to spamc. Some tricks are needed to 
# decide on one user when there are many recipients. This string is of
# course expanded. If unset or empty, spamc will use the user Exim 
# runs as. We suggest that you decide what username to use in the ACLs
# and set an ACL variable.
#SAspamcUser: $acl_m2
```

So it seems my spam checks were performed as Debian-exim due to this variable being unset, and the statuses of the two .spamassassin folders seem to confirm this guess ("Rgp" means August):
```
# ls ~Debian-exim/.spamassassin/ -l
viso 87708
-rw------- 1 Debian-exim Debian-exim 21008384 Lap  1  2014 auto-whitelist
-rw------- 1 Debian-exim Debian-exim        6 Lap  1  2014 auto-whitelist.mutex
-rw------- 1 Debian-exim Debian-exim     6288 Rgp 17 17:07 bayes_journal
-rw------- 1 Debian-exim Debian-exim     1956 Rgp 17 17:06 bayes.mutex
-rw------- 1 Debian-exim Debian-exim 83865600 Rgp 17 16:52 bayes_seen
-rw------- 1 Debian-exim Debian-exim  5332992 Rgp 17 17:06 bayes_toks
-rw-r--r-- 1 Debian-exim Debian-exim     1487 Lap 24  2009 user_prefs
# ls -l ~debian-spamd/.spamassassin/
viso 2140
-rw------- 1 debian-spamd debian-spamd       6 Rgp 22 18:45 bayes.mutex
-rw------- 1 debian-spamd debian-spamd   86016 Rgp 22 18:45 bayes_seen
-rw------- 1 debian-spamd debian-spamd 2633728 Rgp 22 18:45 bayes_toks
drwxr-xr-x 2 debian-spamd debian-spamd    4096 Bir  6  2015 sa-compile.cache
-rw-r--r-- 1 debian-spamd debian-spamd    1869 Rgp 18 18:19 user_prefs
```

I guess I'll just try copying over all files from `~Debian-exim/.spamassassin` to `~debian-spamd/.spamassassin`? :)

Comment (Udera) [inline docs/configure]: > I guess I'll just try copying over all files from ~Debian-exim/.spamassassin to ~debian-spamd/.spamassassin? :)

And restart spamassassin ;-)

Comment (rimas-kudelis) [inline docs/configure]: OK, done. :)

**Comment by rimas-kudelis** [comment]
I've added a new commit to the PR, beware. :)

File: docs/vexim-acl-check-content.conf [hunk 5]
```diff
@@ -1,44 +1,59 @@
...
-        log_message	= This message contains malware ($malware_name)
+  #
+  warn    message     = This message contains malware ($malware_name)
+          malware     = *
+          log_message = This message contains malware ($malware_name)
+
+  # Reject messages with an "X-Spam-Flag: YES" header.
+  #
+  deny    message   = Sender openly considers message as spam ('$regex_match_string' found).
+          condition = ${if bool{$h_x-spam-flag}}
```
Udera: I just checked this setting. There is an error with string expansion, this one worked for me:
`condition = ${if bool{$h_x-spam-flag:}{true}{false}}`
The message should be fixed as well, the variable `$regex_match_string` is not defined.

Udera:
I started testing (only mailbox, no forwards yet). I'm not sure when we should add the `X-Spam-Flag: YES`. We have the `sa_tag` (default: 2) and the `sa_refuse` (default: 5), currently we are adding it for all `score > sa_tag`.

Spamassassin itself uses a threshold of 5. A mail with score 3:
```
Aug 28 23:31:46 freebsd spamd[52421]: spamd: clean message (3.0/5.0) for spamd:58 in 0.5 seconds, 483 bytes. 
```
And if you feed spamassassin the mail:
```
Return-path: <mail@example.org>
X-Spam-Checker-Version: SpamAssassin 3.4.1 (2015-04-28) on freebsd
X-Spam-Level: ***
X-Spam-Status: No, score=3.0 required=5.0 tests=D_TAG autolearn=no
	autolearn_force=no version=3.4.1
....
```

If I take the wording literally:
score > sa_refuse -> mail server refuses mail at delivery
sa_refuse > score > sa_tag -> mail server moves mail in spam folder
score < sa _tag -> mail is delivered

How I use it currently:
score > sa_refuse -> mail server moves mail in spam folder
sa_refuse > score > sa_tag -> mail is delivered, the spamassassin tests are added in the header
score < sa _tag -> mail is delivered

rimas-kudelis:
SpamAssassin threshold of 5 is just a default setting as well. Since our virtual users may have different tag and refuse numbers, and Exim cannot pass any of that information to SpamAssassin, we get to ignore whether or not SA thinks an email is actually spam, and instead have to rely on just the score. So it is quite possible to have `X-Spam-Flag: YES` in combination with `X-Spam-Report: No...` or vice-versa with Vexim, due to the way SA protocol currently works.

About the way things should work with regard to our settings: until moving to Spam folder was introduced, it was easy: `score > sa_refuse`  -> mail was disarded, `score > sa_tag`  -> mail was marked by adding `X-Spam-*` headers. Now the situation is different, and indeed it's not so self-evident anymore. Maybe we should rethink our strategy. One of the options that comes to mind is dropping `sa_refuse` for good, and instead using `sa_tag` as a threshold of delivering the message into the Spam folder. We could also have a boolean setting which would control whether or not `X-Spam-*` headers must be added to all messages. The UI could look somewhat like this:

```
Mark as Spam score                    [________]
   Messages scoring the amount of points specified above or more will be delivered 
   to the user's Spam folder.
   NOTE: regardless of this setting, the server will refuse delivery of messages
   scoring 15 or more points.

Put Spam info headers in all messages [_]
   If checked, X-Spam-Score and other related headers will be present in all messages
   delivered locally. Otherwise, most of these headers will only be present in messages
   considered as spam according to the setting above.
```

All of this doesn't block this PR though, so I suggest we take this part of discussion elsewhere

Udera:
I did some further testing. When you use forwarding addresses, and you activate spam checks on the forwarding address and the mailbox where it is finally delivered, identical headers are added. I tried to set a condition, that headers are only added when not present (def:h_x-spam-flag) but it seems not to work on headers that were added in the router-section.

I haven't tried yet to use own variables in that part. Or we don't add headers when the forwarding address is on the server as well (how are we going to check this, what if there is a chain of forwards?

Udera:
@rimas-kudelis also good ideas. Why using two spam-levels at all. But we can put this in a different issue.

rimas-kudelis:
> I did some further testing. When you use forwarding addresses, and you activate spam checks on the forwarding address and the mailbox where it is finally delivered, identical headers are added. I tried to set a condition, that headers are only added when not present (def:h_x-spam-flag) but it seems not to work on headers that were added in the router-section.
> 
> I haven't tried yet to use own variables in that part. Or we don't add headers when the forwarding address is on the server as well (how are we going to check this, what if there is a chain of forwards?

We could move `headers_add` statements to the transports then. Although given that both headers would be identical, I guess it's not a big issue either way.

rimas-kudelis:
> @rimas-kudelis also good ideas. Why using two spam-levels at all. But we can put this in a different issue.

Another option: keep two levels, but make `sa_tag` mean the level at which the message is moved into the spam folder, and `sa_drop` mean what it meant previously – the level at which the message is discarded. To disable any of these two actions, the user could set the threshold to a ridiculously high value (e.g. 999). Or perhaps zero. Or we could have checkboxes next to them.

Udera:
I'd prefer to reject the message. We discussed this before, so with 
several recipient, we would need to calculate the max of all of the 
recipient sa_refuse-score. We already started to do some exim-magic ...

rimas-kudelis:
Rejecting by the maximum threshold for all users is cool. The bigger question is what to do when the score is less than max, but above user's threshold. We used to drop such messages, and I think it still makes sense to let the user choose to drop them.

Udera:
> We could move headers_add statements to the transports then. Although given that both headers would be identical, I guess it's not a big issue either way.

That would only add X-Spam-* headers to mailboxes and not on forwarded mails.

I found a different solution, I used this router (spam-router would also need a rework):
````
  headers_add = ${if and { \
                    {match{$domain}{$original_domain}} \
                    {match{$local_part}{$original_local_part}} \
                    {>{$spam_score_int}{${extract{sa_tag}{$address_data}}}} \
                    {eq{1}{${extract{on_spamassassin}{$address_data}}}} \
                    } {X-Spam-Flag: YES\nX-Spam-Scoree: $acl_m_spam_score\nX-Spam-Report: $acl_m_spam_report}{} }
````
Now the settings of the original incoming email address are preferred.

rimas-kudelis:
I don't think we want `X-Spam-*` headers in forwarded emails, for exactly same reason as we don't want them in emails we get.

rimas-kudelis:
That said, I still prefer setting these headers in the router, not transport. (y)

Udera:
We could put in a condition that they are only added for local accounts and not for redirects.

Udera:
> I don't think we want X-Spam-* headers in forwarded emails, for exactly same reason as we don't want them in emails we get.

Question is, if it would be interesting to have different spam-limits on local redirects. For external mails we would need to remove all these headers (or block the mail if it is considered spam).

rimas-kudelis:
I've updated the PR. Squashed most commits, added changes to `headers_add`, turned the name of the spam report header into a macro. My editor cleaned up some whitespace in process, I've put these changes into a separate commit.

Udera:
Good. From my POV we can merge it. Or do you want to finish the discussion about the final spamassassin implementation?

rimas-kudelis:
No, I'm good.

PR #205, opened 2016-08-19
Review order of RCPT-ACLs

Udera:
Changes:

* accept authenticated senders before checking blacklists
* put everything optional into `vexim-acl-check-rcpt`

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -343,9 +343,6 @@ acl_check_rcpt:
 
   accept  hosts = :
 
-  # Include Vexim specific rcpt ACLs
-  .include /usr/local/etc/exim/vexim-acl-check-rcpt.conf
-
   # Deny if the local part contains @ or % or / or | or !. These are rarely
   # found in genuine local parts, but are often tried by people looking to
   # circumvent relaying restrictions.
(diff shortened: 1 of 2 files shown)


docs/configure
```diff
@@ -363,22 +360,17 @@ acl_check_rcpt:
...
-  # deny    message       = rejected because $sender_host_address is in a black list at $dnslist_domain\n$dnslist_text
-  #         dnslists      = black.list.example
-  #
-  # warn    message       = X-Warning: $sender_host_address is in a black list at $dnslist_domain
-  #         log_message   = found in $dnslist_domain
-  #         dnslists      = black.list.example
-  #############################################################################
+
+  # Accept if the message arrived over an authenticated connection, from
+  # any host. Again, these messages are usually from MUAs, so recipient
```
rimas-kudelis: What's the "Again" for?

docs/configure
```diff
@@ -407,11 +400,6 @@ acl_check_rcpt:
 
   accept  hosts         = +relay_from_hosts
```
rimas-kudelis: Oh, I see what the "Again" is for. Perhaps the block above should also be moved upwards?

Udera:
Moved the relay-host block upwards. Only thing I'm not sure about, how does exim get the hostname? If it is just taken from the `EHLO` during the SMTP session, this value is not trustworthy. You could guess relay hosts. We should verify this to avoid someone becoming an open relay.

docs/configure
```diff
@@ -363,22 +360,22 @@ acl_check_rcpt:
...
+
+  accept  authenticated = *
+
+  # Accept if the message comes from one of the hosts for which we are an
+  # outgoing relay. Recipient verification is omitted here, because in many
+  # cases the clients are dumb MUAs that don't cope well with SMTP error
+  # responses. If you are actually relaying out from MTAs, you should probably
+  # add recipient verification here.
+
+  accept  hosts         = +relay_from_hosts
```
rimas-kudelis: Just fix this whitespace, and I think this can be merged in.

rimas-kudelis:
DNS, I would guess?

docs/configure
Udera: which one?

Udera:
Yes, perhaps. A reverse DNS lookup on the IP address. Or query the DNS name and check if one of the IPs is identical with the sender. Perhaps exim is already handling this, but we should check this.

docs/configure
rimas-kudelis: The huge one before the `=`

rimas-kudelis: Also, the comments are written in a way where `accept authenticated` goes after `accept hosts`. If you want these conditions in this order, you should update the comments. Although I'd just switch them back to their original order, I don't think it's goin' to make much difference.

Udera: > The huge one before the =

They are all aligned at the same position. Have a look at the following rules, there it is the same.

rimas-kudelis: okay, but the other nit still applies

PR #207, opened 2016-08-23
Prevent XSS attacks

inoveb:
The $_GET variables are echoed unmodified which represents a XSS risk. 
All the pages that require config/header.php are vulnerable.

### DIFF
--- a/vexim/config/functions.php
+++ b/vexim/config/functions.php
@@ -216,4 +216,13 @@
       $_SESSION = array();
       session_destroy();
     }
-?>
+
+    /**
+     * Makes any text safe to be displayed on a web page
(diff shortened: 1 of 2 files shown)


Udera:
Oops, that must be fixed.
Should we also consider setting up a content security policy->http://www.html5rocks.com/en/tutorials/security/content-security-policy/? @rimas-kudelis

rimas-kudelis:
The patch looks good, although I haven't tested.
CSP is a separate issue that can be tackled in another bug/PR.

Udera:
I just checked that error messages are still shown but not all combinations. Looks good for me as well.
CSP will be a new PR.

PR #213, opened 2016-08-30
Support MySQL pre-5.6.1 in the setup script

### LINKED ISSUE #212: mysql.sql doesn't work with Mysql<5.7
Apparently, the TO_BASE64 didn't exist until MySQL 5.7. Since that functionality is not critical, we should probably change it to work with older versions as well.

rimas-kudelis:
The `TO_BASE64()` function was only added in MySQL 5.6.1. This changes the way siteadmin password is generated.
On older versions of MySQL, uppercase-only password will be generated. On 5.6.1 and above, it will be mixed-case as previously.

Fixes #212


Udera:
On mysql 5.5.50 (ubuntu 14.04), I get this error:
````
ERROR 1267 (HY000) at line 218: Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation 'concat'
````

rimas-kudelis:
Could you try again please?

Udera:
Still the same error ;-(

Udera:
This works for this setup:
````
INSERT INTO `users` VALUES (1,1,'siteadmin','siteadmin',ENCRYPT(@SITE_ADMIN_PW,CONCAT(@PW_PREFIX,CONVERT(MD5(RAND()) USING utf8))),65535,65535,'','','site',1,0,0,0,0,0,0,0,1,NULL,NULL,0,0,0,'SiteAdmin',0,0,NULL,NULL);
````

rimas-kudelis:
Thank you, adjusted. Let's see how it goes now.

Udera:
Works now on 5.5.30 (ubuntu 14.04) and 5.6.30 (FreeBSD 10.1).

Issue #211, opened 2016-08-30
Cannot add relay domain


rimas-kudelis:
It seems like #152 has unintended consequences: when I try to create a relay domain, I get the "passwords are too weak" message, even though we aren't adding passwords to this type of domains.

@Udera, @runout-at: which one of you would like to fix this? ;)

rimas-kudelis:
I think the problem is https://github.com/vexim/vexim2/blob/master/vexim/siteaddsubmit.php#L83: Only domains of type `relay` should also be excluded from the check (in other words, it should only be run for `local` domains).

PR #214, opened 2016-08-31
No passwordcheck for relay-domains

Udera:
Fix for #211

`alias`-domains are inserted in a completely different table, therefore the split. Indeed we only need passwords for a new local domain, and there was a default password set. Now I changed it to verify passwords only for local domains.

### DIFF
--- a/vexim/siteaddsubmit.php
+++ b/vexim/siteaddsubmit.php
@@ -72,9 +72,22 @@
     $smtphomepath = $domainpath . "/" . $_POST['localpart'] . "/Maildir";
     $pophomepath = $domainpath . "/" . $_POST['localpart'];
   }
-//Gah. Transactions!! -- GCBirzan
-  if ($_POST['type'] !== "alias") {
-    if ($_POST['type'] === "local") {
+
+  if ($_POST['type'] === "alias") {
+    $query = "INSERT INTO domainalias (domain_id, alias)
+              SELECT domains.domain_id, :alias FROM domains WHERE domains.domain_id=:domain_id";
+    $sth = $dbh->prepare($query);
+    $sth->execute(array(':domain_id'=>$_POST['aliasdest'], ':alias'=>$_POST['domain']));
+    if ($sth->rowCount()!==1) {
+      header ("Location: site.php?failaddeddomerr={$_POST['domain']}");
+      die;
+    } else {
+      header ("Location: site.php?added={$_POST['domain']}" .
+              "&amp;type={$_POST['type']}");
+      die;
+    }
+  } else { // local or relay
+      if ($_POST['type'] === "local") {
         if (!validate_password($_POST['clear'], $_POST['vclear'])) {  
           header ("Location: site.php?badpass={$_POST['domain']}");
           die;
@@ -118,8 +131,6 @@
                 ':pop'=>$pophomepath,
                 ':domain'=>$_POST['domain'],
                 ));
-           
-// Is using indexes worth setting the domain_id by hand? -- GCBirzan
         if (!$success) {
           header ("Location: site.php?failaddedusrerr={$_POST['domain']}");
           die;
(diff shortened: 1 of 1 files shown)


vexim/siteaddsubmit.php
```diff
@@ -79,12 +73,13 @@
...
   }
 //Gah. Transactions!! -- GCBirzan
-  if ((validate_password($_POST['clear'], $_POST['vclear'])) &&
-    ($_POST['type'] != "alias")) {
-      if (!password_strengthcheck($_POST['clear'])) {  
-        header ("Location: site.php?weakpass={$_POST['domain']}");
-        die;
-      }
+  if ($_POST['type'] !== "alias") {
+    if ($_POST['type'] === "local" && validate_password($_POST['clear'], $_POST['vclear'])) {
```
rimas-kudelis: If `validate_password` fails, this will create the domain, without even checking if password is strong.

Line 140 should be changed to simple `else`, and lines 153-154 should be moved below this condition.

Udera:
That's the problem if you want to make it too short and elegant ;-)

rimas-kudelis:
> Line 140 should be changed to simple else, and lines 153-154 should be moved below this condition.

😒

Udera:
I missed that, will do it tonight.

Udera:
Instead of
```
if ($_POST !== "alias") ...
elseif ($_POST === "alias") ...
else ...
```
We just use:
```
if ($_POST === "alias) ...
else ....
```

vexim/siteaddsubmit.php
```diff
@@ -78,13 +72,31 @@
...
+    $query = "INSERT INTO domainalias (domain_id, alias)
+              SELECT domains.domain_id, :alias FROM domains WHERE domains.domain_id=:domain_id";
+    $sth = $dbh->prepare($query);
+    $sth->execute(array(':domain_id'=>$_POST['aliasdest'], ':alias'=>$_POST['domain']));
+    if ($sth->rowCount()!==1) {
+      header ("Location: site.php?failaddeddomerr={$_POST['domain']}");
+      die;
+    } else {
+      header ("Location: site.php?added={$_POST['domain']}" .
+              "&amp;type={$_POST['type']}");
```
rimas-kudelis: Huh, I wonder if we actually need to escape ampersands in HTTP headers.

vexim/siteaddsubmit.php
```diff
@@ -142,22 +152,8 @@
...
-      die;
-    } else {
-      header ("Location: site.php?added={$_POST['domain']}" .
-              "&amp;type={$_POST['type']}");
-      die;
-    }
-  } else {
-    header ("Location: site.php?failaddedpassmismatch={$_POST['domain']}");
   }
+  header ("Location: site.php?failaddedpassmismatch={$_POST['domain']}");
```
rimas-kudelis: This one was supposed to be moved upwards instead of adding the `header ("Location: site.php?badpass={$_POST['domain']}");` line.

vexim/siteaddsubmit.php
Udera: No, that looks wrong.

Udera:
Fixed the error message.

The ampersand in the URL are also at more location. I'll do a separate PR.

vexim/siteaddsubmit.php
Udera: I know, why we put it. There was an error in HTML checks.
https://www.w3.org/TR/html4/appendix/notes.html#h-B.2.2

rimas-kudelis: I think you added it by mistake. `header()` is not HTML and should not be escaped.

vexim/siteaddsubmit.php
```diff
@@ -138,26 +148,10 @@
...
-      header ("Location: site.php?added={$_POST['domain']}" .
-              "&amp;type={$_POST['type']}");
-      die;
-    }
-  } else {
-    header ("Location: site.php?failaddedpassmismatch={$_POST['domain']}");
+    } 
   }
+  header ("Location: site.php?failaddeddomerr={$_POST['domain']}");
+  die;
```
rimas-kudelis: I don't think we can ever reach these two lines. And having `die()` as the last PHP call in a file which is not `include()`d is also rather pointless.

Udera:
There is an if()-loop from line 119 (checking on db-insert for new domain) and the header is reached when this fails (in between is a second db-query creating the postmaster-user).
The `die` is useless.

rimas-kudelis:
Ah, I see now. It would have been more obvious if the last `header()` line was swapped with the preceding closing curly brace though, but this is not necessary.

Anyway, please unescape the ampersand. HTTP headers are NOT same as HTML content, so they need no escaping. The link you gave seems irrelevant in our context.

Udera:
I added the changes.

rimas-kudelis:
Thank you!

PR #220, opened 2016-09-08
Consider local-part-suffix addresses during routing of spam mails

### LINKED ISSUE #137: local_part-suffix in spam-router?
Do we want spam-mails to a local_part_suffix to be filtered as well? Especially these routers:
https://github.com/vexim/vexim2/blob/master/docs/configure#L514-L551

### LINKED ISSUE #219: Ability to reject emails from with specific suffic
In my configuration I have `VEXIM_LOCALPART_SUFFIX = +*` setting, so all emails that go to address which starts with `username+` goes to my `username` mailbox.

Still for some suffixes I started to receive SPAM, I'd like to keep `+` suffix settings, but want to be able to reject emails that specifically are send to e.g. `username+foo@example.com`.

I revised all settings and it seems that it's not possible to achieve with current version of vexim (?)

Udera:
fix #137 
@medikoo

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -547,6 +547,10 @@ ditch_malware:
 		                and users.on_avscan = '1' \
                                 and domains.avscan = '1' \
 		                and users.domain_id=domains.domain_id}}}{1} }} {yes}{no} }
+  .ifdef VEXIM_LOCALPART_SUFFIX
+    local_part_suffix = VEXIM_LOCALPART_SUFFIX
+    local_part_suffix_optional
+  .endif
 
 # spam drop router
 ditch_spam_drop:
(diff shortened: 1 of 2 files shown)


docs/configure
```diff
@@ -585,11 +593,16 @@ ditch_spam:
...
     headers_add = ${if and { \
                      {match{$domain}{$original_domain}} \
                      {match{$local_part}{$original_local_part}} \
                      } {X-Spam-Flag: YES\nX-Spam-Score: $acl_m_spam_score\nVEXIM_SPAM_REPORT_HEADER_NAME: $acl_m_spam_report}{} }
+    .ifdef VEXIM_LOCALPART_SUFFIX
+      local_part_suffix = VEXIM_LOCALPART_SUFFIX
+      local_part_suffix_optional
+    .endif
+    retry_use_local_part
+
```
rimas-kudelis: Pls remove this line and I think we're good.

rimas-kudelis:
`docs/debian-conf.d/router/249_vexim_ditch_routers` needs this change as well.

Udera:
Done.

rimas-kudelis:
thanks!

PR #224, opened 2016-09-14
Treat SA tag and refuse scores properly

### LINKED ISSUE #223: Treat SA tag and refuse scores properly
I think SA tag and refuse scores should be treated as minimum required values, not as maximum unfit values. In other words, `>=`, and not `>`.

rimas-kudelis:
`sa_tag` and `sa_refuse` settings should be treated as minimum values to induce the effect, not as maximum values which do NOT induce it.
For example, if I have `sa_tag=20`, I expect all messages with spam score 2.0 or above tagged, as opposed to only messages above 2.0 being tagged.
Fixes #223

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -553,7 +553,7 @@ ditch_spam_drop:
     driver = redirect
     allow_fail
     data = :blackhole:
-    condition = ${if >{$spam_score_int}{${lookup mysql{select users.sa_refuse * 10 from users,domains \
+    condition = ${if >={$spam_score_int}{${lookup mysql{select users.sa_refuse * 10 from users,domains \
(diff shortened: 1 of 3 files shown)


Udera:
We can do this. But it doesn't change so much... Didn't we have somewhere written, mails above this limit are considered as spam ??

rimas-kudelis:
It's definitely a small change. But I don't think that the current policy is explainedaz–––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––  anywhere in the UI. And my expectations from the wording "Spamassassin refuse score" are that mails with that score are refused.

Udera:
I would expect it like this as well.

Issue #232, opened 2017-01-13
Exim 4.88 does not support demime


ecki:
Since Exim 4.88 (on FreeBSD) the "demime" ACL option is removed. I needed to remove some of the rules in vexim-acl-check-content.conf, is it planned to provide a new temlate for this?

Udera:
Removing is just a workaround. There is now a special acl-list for mime-variables:
http://www.exim.org/exim-html-current/doc/html/spec_html/ch-content_scanning_at_acl_time.html

Rules are a bit longer and more diverse, I found a list with some examples here:
http://wiki.wlug.org.nz/EximMailFilter

Thanks to regexp, we could block potentially dangerous double extensions like *.doc.exe. Or only allow a few exceptions like *.tar.gz. We should test these rule with some standard mail programs and webmailers.

rimas-kudelis:
Ahh, github search didn't tell me that the second filter in `vexim-acl-check-content.conf` also uses demime, so I was under an impression that we're only using this option to filter out emails with MIME errors.

I doubt it's worthy to filter particularly on double extensions though, if we disallow .exe and similar dangerous attachments in general.

PR #234, opened 2017-01-17
Remove old demime-function with acl_smtp_mime-settings.

Udera:
Fixes https://github.com/vexim/vexim2/issues/232.

@ecki please test.

### DIFF
--- a/docs/configure
+++ b/docs/configure
@@ -172,9 +172,11 @@ acl_smtp_data = acl_check_content
 
 acl_smtp_helo = acl_check_helo
 
-# This ACL can be used to refuse certain file extension in ZIP files
+# This ACL checks attached files to comply with certain rules (number of
+# attachments, file name length etc.). You can also block certain file
+# extensions.
 
-#acl_smtp_mime = acl_check_mime
+acl_smtp_mime = acl_check_mime
 
 # This configuration variable defines the virus scanner that is used with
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
@Udera is there any rationale to really reject based on "too many MIME parts", "Excessive line length", "Filename length too long" or "MIME boundary length too long"? All of these situations seem quite unlikely, and don't seem dangerous unless your MUA is prone to buffer overflows. Do these edge cases at least contradict any standards?

Udera:
I didn't find a limit for the number of MIME parts. For the line length, I found this:
https://tools.ietf.org/html/rfc5322#section-2.1.1
but I'm not sure if it applies at this point?

The file name length is limited by the file system and NTFS is imposing the 255 character limit (I think to remember). Other users might want to use longer filenames.

Let's only keep the line length if this RFC applies. I wouldn't object to remove the others.

Udera:
I removed all the additional rules. I linked a document in the exim-wiki if someone needs more protection for vulnerable mail clients. The line length is specified in the RFC, but not for binary content. Probably doesn't apply therefore, but I didn't have time to test this extensively.

I tested the current patch on Exim 4.88 on FreeBSD and Debian 8 (4.84).

@rimas-kudelis any more suggestions?

rimas-kudelis:
I haven't tested this myself, but there's no reason not to merge if you have and it works fine for you.

PR #235, opened 2017-02-26
Add Content Security Policy

### LINKED ISSUE #209: Use CSP to increase security
Follow up of https://github.com/vexim/vexim2/pull/207

> setting up a content security policy->http://www.html5rocks.com/en/tutorials/security/content-security-policy/?

Udera:
Fix #209

### DIFF
--- a/vexim/config/httpheaders.php
+++ b/vexim/config/httpheaders.php
@@ -2,4 +2,5 @@
   session_start();
   header("Cache-control: private"); // IE6 hack to back forms + BACK button work
   header("Content-Type: text/html; charset=utf-8");
+  if (isset($CSPenabled) && $CSPenabled === true)  header("Content-Security-Policy: default-src 'self';");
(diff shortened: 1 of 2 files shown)


rimas-kudelis:
I think we should make this configurable, via `variables.php`.
E.g. `$allowframing = false;` by default or something like that.

Udera:
There is a problem with inline-code (java-script/css) that is not supported by default:
https://developer.chrome.com/extensions/contentSecurityPolicy#JSExecution

We have some pages where we use `onload` to focus on the first form field, we can replace this by putting `autofocus` into the field which is supposed to be focused (supported by HTML 5). Another example is the automatic password generator and a few cases with inline css code.

Not sure if it is worth working over all these examples to make CSP work.

Udera:
This option is now configurable and disabled by default. PR updated.

PR #236, opened 2017-04-03
Add Doctype in each file, clean up index.php

Udera:
Put `<!DOCTYPE html>` to each file.

index.php:
- use new `autofocus` option on first field instead of inline-javascript (localpart field doesn't exist anymore https://github.com/vexim/vexim2/pull/110)
- use existing css class instead of inline-css

### DIFF
--- a/vexim/admin.php
+++ b/vexim/admin.php
@@ -4,6 +4,7 @@
   include_once dirname(__FILE__) . '/config/functions.php';
   include_once dirname(__FILE__) . '/config/httpheaders.php';
 ?>
+<!DOCTYPE html>
 <html>
   <head>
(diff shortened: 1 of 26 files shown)


vexim/config/httpheaders.php
```diff
@@ -2,4 +2,5 @@
   session_start();
   header("Cache-control: private"); // IE6 hack to back forms + BACK button work
   header("Content-Type: text/html; charset=utf-8");
+  print '<!DOCTYPE html>';
```
rimas-kudelis: We have files which include `httpheaders.php`, but issue further `header()` calls. If you want this doctype, you should add it before the `<html>` tag to the files that have that tag.

Udera:
A bit of shell scripting later ...

Issue #191, opened 2016-08-16
Update translation


Udera:
Vexim2.3 brings a number of new strings. The existing translations and `vexim2/vexim/locale/template.pot` have to be reviewed.

Udera:
For the new release, we would appreciate if you updated your language locale:
@soujak @kvsmirnov

PR #241, opened 2017-06-07
Some changes in translation

kvsmirnov:
Corrected and updated Russian translation
Added some not translated strings
Merged lines with difference in punctuation or letter case (spam and
Spam, MB and Mb for example)


rimas-kudelis:
I'm not sure if it's worthwhile to make string changes in this old codebase. @Udera what do you think?

Udera:
We asked for some updates: https://github.com/vexim/vexim2/issues/191
But we do need to touch all the other translation files as well?

There are different writings of `e-mail` `email` `mail` ...

rimas-kudelis:
@kvsmirnov please don't put all your fixes and changes in one PR, it will only make it less likely to be accepted. The smaller the updates, the better.

Particularly, I don't think we should accept your changes which add backticks to SQL queries, because that makes them totally incompatible with PostgreSQL. Don't forget we're using PDO – a DB abstraction layer which should be taking care of adding them when using MySQL.

kvsmirnov:
I have MySQL 8 on my FreeBSD servers. MySQL 8 does not accept queries without backticks... Sorry, I have not tested this with PostgreSQL. I will stop using vexim2 in some weeks. My team writing own interface

rimas-kudelis:
I'll say it differently: were you fixing the backticks preemtively, or were you actually affected by the bug? Like I said, PDO is an abstraction layer, so expect it to be adding backticks automatically when querying MySQL.

kvsmirnov:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html

kvsmirnov:
I have MySQL 8.0.0, Apache 2.4, PHP 7.1 at my servers

Issue #30, opened 2015-01-24
Feature request: Specify sending IP address per domain


ConneXNL:
It would be awesome if you could set the sending IP address per domain.

I think this can be achieved relatively easy by setting the "interface" under "remote_smtp" as described in:
https://loginroot.com/how-to-sending-emails-using-user-assigned-ip-in-exim-directadmin/

Udera:
Exim4U provides some additional functions compared to vexim.

Z3po:
I couldn't find anything I currently miss with vexim in exim4u featurelist.... I don't think vexim is dead, it just took some time off :)
Time to come back, isn't it?

rimas-kudelis:
Patches for this feature are welcome. :) From looking at the howto, it seems a rather easy thing to implement. However, couple considerations should be made:
1) most important – domains without outbound IP specified should still work
2) less important – domains with incorrect IP address specified (i.e. some IP which the server is not using at the moment). I'm still unsure whether these should keep working or – on the contrary – stop working. This is open for discussion.

runout-at:
On starting to use IPv6 i had to set the outgoing address because i have more than one IPv6 on the host and exim used the wrong one.
My hack is to set a variable:
in /etc/exim4/conf.d/main/00_vexim_listmacrosdefs
`
OUTGOING_IP_ADDRESSES = y.y.y.y ; 2002::xxxx
`
and in 2 transports:
/etc/exim4/conf.d/transport/30_exim4-config_remote_smtp
/etc/exim4/conf.d/transport/30_exim4-config_remote_smtp_smarthost
```
.ifdef OUTGOING_IP_ADDRESSES
  interface = <; OUTGOING_IP_ADDRESSES
.endif
```

it would be just another field in the domain table and a query instead the fixed variable  (varchar 255?).
I would volunteer to prepare a PR for that.

1. What would you prefer as fallback if a domain has no IP set?
2. How do we check if the IP is not a correct IP for the host or domain? Is this possible by MX lookups?
3. As i modified the original exim transport files i'd be interested in a different solution without touching them. Its a hassle on later Debian upgrades if we change those. Ideas anyone?

Other/more ideas?

Udera:
> What would you prefer as fallback if a domain has no IP set?

I'm not really sure what your use case is. Is just a few domains and all the others work on the default IP?


> How do we check if the IP is not a correct IP for the host or domain? Is this possible by MX lookups?

[...]

PR #245, opened 2017-10-12
outgoing IP per domain

runout-at:
this is still missing the part for the webUI.
But i want to discuss if this is going in the right direction.

#30


Udera:
You are planning that you can put a list of IPs? Or how do you manage ipv4 and ipv6?

runout-at:
The easiest solution would be to have just one additional field in the 'domains' table and put a semicolon separated list of IPs there. This way we can have mix of IPv4 and IPv6 in this field and can use it 'as is' in the transport like `interface = <; OUTGOING_IP_ADDRESSES`

Maybe the WebUI could be more complex if we like to check/sanitize the values on input or split ddresses in separate fields - but to me, one field would be enough. I wouldn't do checks on every mail in exim.
One could monitor the validity of the addresses with a cron job.

runout-at:
This is on my production servers for 2 weeks without issues, IPv4/v6 dual stack.

The only part missing here is the transports.
For now i add following to the original Debian transport files but i'm not sure what is the best way to integrate this in vexim:
```
.ifdef OUTGOING_IP_ADDRESSES
  interface = <; OUTGOING_IP_ADDRESSES
.endif
```
added to following files:
/etc/exim4/conf.d/transport/30_exim4-config_remote_smtp
/etc/exim4/conf.d/transport/30_exim4-config_remote_smtp_smarthost

Udera:
Let's wait for the 2.4 milestone when the new webinterface is planned. Though it seems very far away.

runout-at:
i think that we can have the actual changes earlier and do the webUI later.

Having different IPs per domain is a good feature. Besides from IPv6 it makes it easier getting mails through spam filters at gmail and others - eg if some domains on a server are sending bulk mails, newsletters,... it's better other mail is going out over a different IP.

runout-at:
A rework of this PR makes sense since Debian has integrated this in the default configs now.

I did use the variable _OUTGOING_IP_ADDRESSES_. Debian uses _REMOTE_SMTP_INTERFACE_. That's the only difference.

We only really need the changes in the DB and a config setting to read the IP from the database like:

- /etc/exim4/conf.d/main/00_vexim_listmacrosdefs

```
VEXIM_OUTIP = SELECT DISTINCT out_ip FROM domains \
                           WHERE domain = '${quote_mysql:$sender_address_domain}' \
                               AND out_ip <> TRIM('')
OUTGOING_IP_DEFAULT = 2001:db8:: ; 10.0.0.1
REMOTE_SMTP_INTERFACE = ${lookup mysql{VEXIM_OUTIP}{ <; $value }{ <; OUTGOING_IP_DEFAULT }}
```

Udera:
How does the smtp transport know which ip to use (or is this already hidden somewhere)? You just add all of the ip addresses:

> The first interface of the correct type (IPv4 or IPv6) is used for the outgoing connection.

https://www.exim.org/exim-html-current/doc/html/spec_html/ch-the_smtp_transport.html

runout-at:
In Debian split config its in transport/30_exim4-config_remote_smtp

```
.ifdef REMOTE_SMTP_INTERFACE
  interface = REMOTE_SMTP_INTERFACE
.endif
```

My code from above does set the variable _REMOTE_SMTP_INTERFACE_ from the DB-query or a default value.

runout-at:
After reading your comment again, I'm not sure I did understand your question right. Maybe my answer was not sufficient.

If multiple IPs are given in _REMOTE_SMTP_INTERFACE_, Exim will try each until it finds one that is sufficient. Eg if the connection is IPv4 and the first addresses given are IPv6 it will skip it. At least it works like that on my production sever for several years now.

PR #248, opened 2017-11-25
Postmaster no delete

### LINKED ISSUE #228: Always accept postmaster email
I wonder what currently happens if we accept an e-mail to postmaster for a virtual domain which doesn't have a postmaster mailbox/forward. When I fix #218, we'll probably start rejecting such e-mails instead. Since RFC's require us to have such address, I'm thinking what options we have.

1. We could create the postmaster address automatically, and not allow deleting it. This would however make it impossible to turn it into a forward using conventional means.
2. We could redirect all our postmaster email to the main postmaster address on localhost, or at least fall back to that.
3. We could test if the postmaster address exists when domain admin logs in, and if it doesn't, display a big fat red warning on top of the page. This would act as a reminder, but wouldn't be strictly enforcing. Although we could also disallow creation of any other mailboxes/aliases until postmaster is created.
4. Combine 3 and 2 ?

Udera:
Fixes https://github.com/vexim/vexim2/issues/228

I have created two commits:
- Do not allow to delete postmaster account (don't show delete button, and even if the id is given manually to the deletion script, the postmaster account is not deleted).
- If no postmaster-account exists, the mails are automatically forwarded to all admin accounts. This is a fallback for setups without postmaster account.


runout-at (approved):

rimas-kudelis:
Hmmm.... If we enable automatic postmaster forwarding to admins, then there's probably no reason to disallow deleting the postmaster account, or is there?

rimas-kudelis:
Also, does the new router actually end up being used? If so, is it being used only for postmaster account?

Sorry, I've been keeping a blind eye on this project for a while, so these questions may be a bit silly.

vexim/adminuserdelete.php
```diff
@@ -4,12 +4,18 @@
   include_once dirname(__FILE__) . '/config/functions.php';
   include_once dirname(__FILE__) . '/config/httpheaders.php';
 
+# Do not allow to delete the postmaster (RFC5321)
+if ($_GET['localpart']==='postmaster') {
```
rimas-kudelis, line 8: While this approach should work, don't you think it would be better to just fetch the record by id from the DB and check if its localpart is not `postmaster` instead?

rimas-kudelis (changes requested):

Issue #246, opened 2017-10-29
SSL certificates per Domain


runout-at:
The idea is that we have different SSL certs. Eg. one per domain.
Normally i get my certificates with certbot from letsencrypt.
For some people it's confusing to see a certificate from a different domain on connecting to the mailserver.

I don't know if or how easily this could be done.
Let me know if you think this is completely stupid.

Udera:
Wouldn't you need a dedicated IP address for each certificate?

runout-at:
SNI should work with SMTP the same way as on HTTPS.
http://www.exim.org/exim-html-current/doc/html/spec_html/ch-encrypted_smtp_connections_using_tlsssl.html#SECTtlssni

runout-at:
With SNI it should work.
An example:
https://help.atmail.com/hc/en-us/articles/115009208748-Multiple-SSL-certs-with-Dovecot-Exim
Exim docu:
http://www.exim.org/exim-html-current/doc/html/spec_html/ch-encrypted_smtp_connections_using_tlsssl.html#SECTtlssni

Udera:
Do you know which clients work well with SNI? Other mail servers (workaround: you could use SNI only on port 587/465)?
https://wiki.dovecot.org/SSL/SNIClientSupport
https://wiki.gentoo.org/wiki/Complete_Virtual_Mail_Server/SSL_Certificates
http://postfix.1071664.n5.nabble.com/TLS-SNI-support-td72278.html

Since letsencrypt allow multidomain-certs, it's only a cosmetic thing if someone checks the certificate manually (how do you check the ssl cert in a mail application?). For browser access you could still use a subdomain (webmail.example.org) using SNI.

runout-at:
If the client is not sending an SNI hostname, then there is a 'fallback' to a certificate which has all hostnames in the SAN field. Or you could see it as a 'default' as long as the client doesn't send SNI.
This way there should be no problem with any client.
Mailservers (on port 25) normally don't send SNI - the only SNI i saw was from google.
(I'm running this setup on my production server now)

Actually the **only really needed config for exim is in the 2 lines** from my PR and this has no negative impact if there is only one certificate using SANs.
The **bash script is just for making it easier to deploy the certificates** based on the records in the vexim-DB.
This script is also doing renewal of the certificates - put it in a daily cron job doesn't hurt.

[...]

PR #249, opened 2017-12-01
Exim multi cert

runout-at:
support for multiple ssl-certificates
#246

### DIFF
--- a/docs/mon-certificates
+++ b/docs/mon-certificates
@@ -8,11 +8,7 @@
 # Certbot will listen on the IP resolved by DNS
 # and the ACME server will use the same IP for validating
 #
-# make sure the host will be reachable by the IPs which
-# will be resolved by DNS for the domainnames the MX records
-#
-# renewal will be done only for certificates close to expiry
-# this is default beahaviour of certbot
+# tested with debian stretch, letsencrypt, exim4, dovecot, nginx
 ###########
 
 # whitespace separated list of subdomain prefixes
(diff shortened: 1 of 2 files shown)


Udera:
Nice work! So, it's mainly for the user that they can use their own domain name for all mail-related services which is great. And if you are concerned about SNI-support in other mail servers (for incoming mail) you still ca use the default hostname in the mx-entry of the DNS. I think it's a very nice feature for vexim.

I didn't go through the code yet. It's probably for debian systems, I will see if I can get this working under FreeBSD.

metacreo:
I don't understand something. It works on my servers without any scripts. What is all this for?
certbot renew automatically all cert`s
just added:
`tls_certificate = ${if exists{CERT_PATH/${tls_sni}/fullchain.pem}{CERT_PATH/${tls_sni}/fullchain.pem}{CERT_PATH/${primary_hostname}/fullchain.pem}}
tls_privatekey = ${if exists{CERT_PATH/${tls_sni}/privkey.pem}{CERT_PATH/${tls_sni}/privkey.pem}{CERT_PATH/${primary_hostname}/privkey.pem}}
`
to exim cfg and written daemons reload in certbot hook

runout-at:
the script 'mon-certificates' makes sure that all certificates in the database are getting generated by certbot. just tin case you are lazy and wouldn't do it manually.

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+
+# set this if a config file for dovecot should be created
+# this will also reload dovecot
+# will be used only if the dovecot binary is executable
+DOVECOT_SSL="/etc/dovecot/conf.d/11-ssl-domlist.conf"
+DOVECOT_BIN="/usr/sbin/dovecot"
+
+# set this if a config file for dovecot should be created
+# this will also reload dovecot
+# will be used only if the dovecot binary is executable
```
rimas-kudelis, line 66: This says "dovecot", but obviously Exim is meant here.

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+#NGINX_SSL="/etc/nginx/sites-available/vexim"
+#NGINX_BIN="/usr/sbin/nginx"
+
+# settings for the nginx config file
+# where should nginx pass the request. e.g.:
+# proxy_pass http://mail1;
+# uwsgi_pass 0.0.0.0:8000;
+NGINX_PASS="proxy_pass http://mail1;"
+# a file where additional parameters for proxy configurations will be
+NGINX_PROXYPARAMS="include /etc/nginx/proxy_params;"
```
rimas-kudelis, line 83: `NGINX_PROXY_PASS` and `NGINX_PROXY_PARAMS` would look nicer

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+# and smtp.example.com
+# additionally certificates for all DNS MX records of example.com
+# will be requested automagically
+# no certificate for example.com itself will be requested
+SUBDOMAIN_LIST_DOVECOT="mail"
+SUBDOMAIN_LIST_EXIM="smtp"
+SUBDOMAIN_LIST_WWW="vexim"
+
+# if you want to use a dedicated DNS resolver put it here
+DNSSERVER="ns1.runout.at"
```
rimas-kudelis, line 25: Do you really want to leave your own DNS server as example?

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+# type of IP address should be used "AAAA" or "A"
+DNSTYPE="AAAA"
+
+# If you're behind a nat or you have some kind of port forwarding
+# you can set a global IP for certbot here.
+# certbot will listen on this IP but the ACME server will use the IP from DNS
+DOMAIN_IP_OVERRIDE=""
```
rimas-kudelis, line 32: Options that are optional should probably be commented out?

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+# will be requested automagically
+# no certificate for example.com itself will be requested
+SUBDOMAIN_LIST_DOVECOT="mail"
+SUBDOMAIN_LIST_EXIM="smtp"
+SUBDOMAIN_LIST_WWW="vexim"
+
+# if you want to use a dedicated DNS resolver put it here
+DNSSERVER="ns1.runout.at"
+# type of IP address should be used "AAAA" or "A"
+DNSTYPE="AAAA"
```
rimas-kudelis, line 27: This lacks an explanation about what record we are talking about here.

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
...
+    mv -f ${DOVECOT_SSL} ${DOVECOT_SSL}.1 2>/dev/null
+    SUBDOMAIN_LIST_DOVECOT="$(echo -e "${SUBDOMAIN_LIST_DOVECOT}" | sed "s/\s/\n/g" | sed '/^[[:space:]]*$/d')"
+    echo -e "DOVECOT config file: ${DOVECOT_SSL}" >> "${LOG_FILE}"
+else # unset DOVECOT_SSL if no executable binary is found
+    DOVECOT_SSL=
+    SUBDOMAIN_LIST_DOVECOT=""
+fi
+
+if [ ! -z "${EXIM_SSL}" ] && [ -x "${EXIM_BIN}" ]; then
+    mv -f ${EXIM_SSL} ${DOVECOT_SSL}.1 2>/dev/null
```
rimas-kudelis, line 116: Apparent copy-paste error here.

docs/mon-certificates
```diff
@@ -0,0 +1,288 @@
+#!/bin/bash
```
rimas-kudelis, line 1: `/bin/sh` would be more portable.

rimas-kudelis (commented):
I tried running the script locally and it failed. Maybe that's because I wasn't running it as root.
I got a bunch of lines like this in the output:

> dig: '.admin' is not a legal name (empty label) # in other lines `.admin` was replaced with my TLD's.

Also I had a bunch of other comments, see below.

Also, there's currently no place in our distribution dedicated to scripts like this, so I'm reluctant to merge this in. Maybe a better place for it would be in the Wiki, at least for now?

rimas-kudelis (changes requested):

runout-at:
thx for the feedback. The script is running on my server daily. I'll check if I have made changes and report back.

This would be the interesting part for a merge:

[‎docs/debian-conf.d/main/00_vexim_listmacrosdefs](https://github.com/vexim/vexim2/pull/249/changes#diff-36b5071f8262d0f03e1071d661311c035b7a520ca3990ebd43464ed31f251479)

Maybe I should have made separate PRs for the monitoring scripts.

rimas-kudelis:
The macros change also seems quite specific to your server. First, because it's probably Debian-specific, but also because it uses `/DEFAULT/` as fallback, which I suspect is a symlink you might have created yourself?

PR #250, opened 2017-12-23
Make tables sortable with js

Udera:
The FreeBSD port (https://www.freshports.org/mail/vexim/) has a patch makes the user tables sortable by mail address instead of mail alias:
https://svnweb.freebsd.org/ports/head/mail/vexim/files/extrapatch-order?revision=435583&view=markup

Having the tables sortable could be very useful, I found this js which allows us to do this, implementation is quite easy: https://kryogenix.org/code/browser/sorttable/

### DIFF
--- a/vexim/adminalias.php
+++ b/vexim/adminalias.php
@@ -8,6 +8,7 @@
 <html>
   <head>
     <title><?php echo _('Virtual Exim') . ': ' . _('Manage Users'); ?></title>
+    <script src="sorttable.js"></script>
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
Is this tested and and does it work? The HTML overhead seems really minimal, so I guess why not.

Udera:
I just tested it on my system. So it would be great if somebody else could test this as well.

rimas-kudelis (approved):

Issue #242, opened 2017-06-14
Migrate MySQL database to `utf8mb4` (or ASCII where appropriate)


runout-at:
With following subject line i get `R=ditch_hdrsubject defer (-1): condition check lookup defer`

`Subject: =?utf-8?Q?Fwd:_=E2=9C=94_Viel_Spa=C3=9F_beim_Erleben!?=`

As far as i can interpret this issue, there is a problem with the lookup in the database with the UTF8 char `=E2=9C=94` in the file `/etc/exim4/conf.d/router/249_vexim_ditch_routers`

For now i put `condition = no` in this router but this is _not_ a solution.

Maybe this problem can also arise on other DB-lookups.
I'm not sure if this can be related to the Mysql/Mariadb-UTF8 bug and a solution could be to move the database to UTF8mb4 encoding. As i have a lot of DBs on my installation it would be hard to change the charset.

rimas-kudelis:
Hi!
Could you try running `exim4 -d -v your@email.address` as root, and then submitting a simple message to the stdin, something like this:
```From: nobody@example.org
To: your@email.address
Subject: =?utf-8?Q?Fwd:_=E2=9C=94_Viel_Spa=C3=9F_beim_Erleben!?=

Test email
```
And press Ctrl+D to signal end of input.

I wonder what Exim would tell you on the standard output, in the `--------> ditch_hdrsubject router <--------` section. For me, I don't see anything suspicious, but I think I remember MySQL becoming more picky about character sets lately, so it might be that. Either way, it would be nice to see the actual MySQL error code or message to debug this properly.

For the reference, my database (and particularly the tables affected) is utf8, not utf8mb4.

runout-at:
Sorry for my delay, have almost no internet until mid of september and i don't want to mess around with a production machine in this (remote) situation.

runout-at:
This seems to happen in `ditch_hdrto` too:
To: =?UTF-8?Q?Gy=c5=91XXX_XXXXX?= <XXXXX.XXXXX@XXXXX.XX>

i hope to find time in the next days to look deeper into this.

Udera:
I think I tested on my setup with `utf8_general_ci` and didn't have a problem.
Can you give more details about your system in order to reproduce this problem?

runout-at:
latin1_swedish_ci  ... it seems that has never changed since the first install of vexim - a long time ago.
Most databases i have converted years ago and i don't know why not the vexim-db.
I should do this anyways...
[...]

PR #253, opened 2018-04-12
fix mysql: raw format, utf8mb4

runout-at:
issue #242


rimas-kudelis:
- [ ] We should also add a migration for this.

But this is a schema change, which I wanted to avoid. But it makes sense, AAAAH! Shall we release 2.4 instead of 2.3.1 then?

runout-at:
After a MySQL Upgrade last summer we'd run into a problem reimporting a database (on the same machine!). The problem was, that the new MySQL did export as utf8mb3 (yes 3!). Just be aware of that could be an issue if you run into a problem.

PR #257, opened 2018-05-05
change is_public in groups table to tinyint

### LINKED ISSUE #256: group feature, is_public should be boolean
The `groups` table has a field `is_public` which is of type string

Since all other Checkbox fields in vexim are boolean (integer in DB) i'd like to change this.
Involved is one router.

runout-at:
To have consistent fieldtypes in vexim
fixes #256

### DIFF
--- a/setup/pgsql.sql
+++ b/setup/pgsql.sql
@@ -168,7 +168,7 @@ CREATE TABLE "groups" (
   "id" SERIAL PRIMARY KEY,
   "domain_id" int NOT NULL,
   "name" varchar(64) NOT NULL,
-  "is_public" char(1) NOT NULL DEFAULT 'Y',
+  "is_public" smallint NOT NULL DEFAULT '1' CHECK("enabled" BETWEEN 0 AND 1),
   "enabled" smallint NOT NULL DEFAULT '1' CHECK("enabled" BETWEEN 0 AND 1),
   UNIQUE("domain_id","name"));
 CREATE INDEX "fk_groups_domain_id_idx" ON "groups" ("domain_id");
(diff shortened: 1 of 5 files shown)


setup/pgsql.sql
```diff
@@ -168,7 +168,7 @@ CREATE TABLE "groups" (
   "id" SERIAL PRIMARY KEY,
   "domain_id" int NOT NULL,
   "name" varchar(64) NOT NULL,
-  "is_public" char(1) NOT NULL DEFAULT 'Y',
+  "is_public" smallint NOT NULL DEFAULT '1' CHECK("enabled" BETWEEN 0 AND 1),
```
rimas-kudelis, line 171: You forgot to adjust the text you copied. :)

rimas-kudelis (commented):
Also, this will need a migration.

I guess I'm ready to shed my reluctance to change our DB schema.

setup/pgsql.sql
runout-at: sorry. I'm not so familiar with psql

rimas-kudelis: I'm guessing that `"enabled"` is the name of the column in the line below, so this `CHECK` should mention `"is_public"` instead.

runout-at: You are right it should be "is_public" on this line.
Somehow github redirects me to a 404-page when I try to edit this file.

rimas-kudelis: Well, you've deleted your branch, hence the 404.

I've [recreated it for you](https://github.com/vexim/vexim2/tree/runout-at-groups-boolfield) and fixed this line. You should be able to change the source of this PR to that branch.

rimas-kudelis:
Closing because source branch cannot be changed(?)

PR #262, opened 2019-11-27
Fix errors when the user is a member of multiple groups

luqelinux:
This query returns multiple records when the user is a member of multiple groups therefore email gets rejected. Exim expects a single record here.


rimas-kudelis:
Thanks @luqelinux !

PR #267, opened 2020-05-06
CLI for Vexim2

ezbik:
I have added some CLI scripts for Vexim2:

vexim2_changepass
vexim2_del_domain
vexim2_del_user
vexim2_list_domainaliases
vexim2_list_users
vexim2_new_alias
vexim2_new_alias_domain
vexim2_new_catchall
vexim2_new_domain
vexim2_new_user


runout-at:
I just found this pull request (sorry, I don't know why I have not seen it earlier)

I still didn't  read it in detail, but my first impression is that it requires mysql and dovecot.
All other scripts of Vexim are more generally. At least psql and courier should be working too.

rimas-kudelis:
I'd be in favor of a single script that does everything depending on the options and arguments passed to it, not a bunch of single-purpose scripts with underscores in their name.

I think Vexim2 actually evolved from a script like that (I remember there was some Perl-based CLI before I switched to Vexim2 in 200x).

Also: new framework-backed frontends might provide their own CLI scripts which do the same thing.

All that said, I think this PR should be closed without merging. We could link to it from the Wiki for those who want something really simple that is already implemented.

PR #269, opened 2020-09-30
Domains table name patch

omatosan:
With typo in `domains` table name in SELECT the check for maxaccounts never happens.


rimas-kudelis:
Thanks!

PR #273, opened 2021-10-28
Fix query to check user group membership

tavil85:
This query does not check if the user is a member of the group. Instead
it returns the username for each group it belongs to.


runout-at:
It seems this will be solved by #285 too. Can someone confirm this. The rewrite uses a different approach for the complete query.

rimas-kudelis:
Let's be realistic: #285 might not happen anytime soon, or at all. If we have errors in our SQL syntax and PRs which fix those errors, I think we should merge the PRs.

I probably haven't merged this in because I'm not using groups and was too lazy to set some up just to test that this PR does what it says it does.

rimas-kudelis:
@tavil85, thanks for the patch and sorry for taking so long!

PR #289, opened 2024-12-13
Fix userchangesubmit.php

metacreo:
Fix type comparison. Got int from db. So... this condition never runs.

`$row['avscan'] === '1'`
But `$row['avscan']` is integer.

Here is 2 ways to fix
`$row['avscan'] === 1` or `$row['avscan'] == '1'`

### DIFF
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -35,8 +35,8 @@
   $sth = $dbh->prepare($query);
   $sth->execute(array(':domain_id' => $_SESSION['domain_id']));
   $row = $sth->fetch();
-  $_POST['on_avscan'] = ($_POST['on_avscan'] ?? null && $row['avscan'] == '1') ? 1 : 0;
-  $_POST['on_spamassassin'] = ($_POST['on_spamassassin'] ?? null && $row['spamassassin'] == '1') ? 1 : 0;
+  $_POST['on_avscan'] = ($_POST['on_avscan'] ?? false && $row['avscan'] == '1') ? 1 : 0;
+  $_POST['on_spamassassin'] = ($_POST['on_spamassassin'] ?? false && $row['spamassassin'] == '1') ? 1 : 0;
   if (isset($_POST['maxmsgsize']) && $row['maxmsgsize'] !== '0') {
     if ($_POST['maxmsgsize'] <= 0 || $_POST['maxmsgsize'] > $row['maxmsgsize']) {
       $_POST['maxmsgsize'] = $row['maxmsgsize'];
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -35,8 +35,8 @@
   $sth = $dbh->prepare($query);
   $sth->execute(array(':domain_id'=>$_SESSION['domain_id']));
   $row = $sth->fetch();
-  if ((isset($_POST['on_avscan'])) && ($row['avscan'] === '1')) {$_POST['on_avscan'] = 1;} else {$_POST['on_avscan'] = 0;}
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
May I suggest the second approach (compare using `==` instead of `===`? I'm not entirely sure we're always getting integers from the database.

If we ever rewrite the webapp using a modern PHP framework like Symfony, we can switch to equivalence checking then.

metacreo:
@rimas-kudelis  Yea sure we can use ==
P/S I make small rewrite of code ... optimized.

vexim/userchangesubmit.php
```diff
@@ -33,16 +33,16 @@
...
-  $sth->execute(array(':domain_id'=>$_SESSION['domain_id']));
+  $sth->execute(array(':domain_id' => $_SESSION['domain_id']));
   $row = $sth->fetch();
-  if ((isset($_POST['on_avscan'])) && ($row['avscan'] === '1')) {$_POST['on_avscan'] = 1;} else {$_POST['on_avscan'] = 0;}
-  if ((isset($_POST['on_spamassassin'])) && ($row['spamassassin'] === '1')) {$_POST['on_spamassassin'] = 1;} else {$_POST['on_spamassassin'] = 0;}
-  if (isset($_POST['maxmsgsize']) && $row['maxmsgsize']!=='0') {
-    if ($_POST['maxmsgsize']<=0 || $_POST['maxmsgsize']>$row['maxmsgsize']) {
-      $_POST['maxmsgsize']=$row['maxmsgsize'];
+  $_POST['on_avscan'] = ($_POST['on_avscan'] ?? null && $row['avscan'] == '1') ? 1 : 0;
+  $_POST['on_avscan'] = ($_POST['on_spamassassin'] ?? null && $row['spamassassin'] == '1') ? 1 : 0;
```
rimas-kudelis: - Copy-paste error here.
- I would put the part before `&&` inside parentheses. Not sure what has precedence and it's IMO just easier to use them than to refer to the docs or try to remember.
- Also not sure why you'd coalesce to `null` instead of e.g. `false`, since you're treating these values as booleans anyway.

rimas-kudelis: @metacreo ↑ at least the second point still applies

metacreo: I think you right false will be better here for future compatibility.

rimas-kudelis: @metacreo,

> Parentheses not needed, here run strong logic only.

Let's see:

```php
php > var_dump(null ?? null && true);
bool(false)
php > var_dump('1' ?? null && true);
string(1) "1"
php > var_dump('1' ?? null && false);
string(1) "1"
```
while the first two cases are somewhat acceptable, that latter one isn't. `$_POST['on_avscan']` should only be set to `1` if both `$_POST['on_avscan']` and `$row['avscan']` are truthy. Just look at the original code, where the only problem is `===` instead of `==`:

```php
if ((isset($_POST['on_spamassassin'])) && ($row['spamassassin'] === '1'))
```

`'1' ?? null && false` works as `'1' ?? (null && false)`, not as `('1' ?? null) && false` as is (or at least should be) the intention here.

This is exactly why the parentheses are needed, and why I sometimes prefer to have them even if they are not needed. They just make logic more explicit and reduce chances for bugs.

> null because it better for if() condition. 

Nah, actually, `false` is better for the `if()` condition, because `null` will need conversion (done internally by PHP). It doesn't matter in terms of speed (not to us, at least), I just think it's "cleaner" that way.

I get the irony of talking about cleanliness in the context of Vexim code, which is why I'm mostly concerned with the second bullet point.


> $_POST['on_spamassassin'] will be processed on positive value only. summary we have here condition `if ([null or (something not goes from post] && $row['avscan'] == '1' )`

I'm not sure I follow what you mean here but my point is that this logic is not correct even if it works as you actually intended it to. :smile:

> `if (null && true/false)` depends on true/false only

No, `if (null && true/false)` is always `false` because PHP considers `null` a falsy value and will treat it as false.

PR #290, opened 2024-12-13
Update userchangesubmit.php

metacreo:
Fix $_POST on_avscan and on_spamassassin checks condition on user save settings.
On my install with mysql 8 and php82 User save settings is incomplete.
Unavailable to save on_avscan and on_spamassassin $_POST params, because getting `int` from db, but checking condition in code accept `string` only.
This commit is fix checking conditions `$value === '1'` to simplest `$value == '1'`

### DIFF
--- a/vexim/userchangesubmit.php
+++ b/vexim/userchangesubmit.php
@@ -35,8 +35,8 @@
   $sth = $dbh->prepare($query);
   $sth->execute(array(':domain_id'=>$_SESSION['domain_id']));
   $row = $sth->fetch();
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
One last thing: would you mind rewriting the commit message to something more informative? Thanks!

metacreo:
> One last thing: would you mind rewriting the commit message to something more informative? Thanks!

Pataisyta. Ar taip bus ok?

rimas-kudelis:
Nepastebėjau/nesulaukiau pakeitimo, tad pats pataisiau. :)

PR #293, opened 2025-02-05
endpoint for user change password

Zetanova:
let user change there password over roundcube webmail

roundcube password plugin `plugins/password/config.inc.php`:
```
    <?php
    $config['password_driver'] = 'httpapi';
    $config['password_confirm_current'] = true;
    $config['password_minimum_length'] = 8;
    $config['password_strength_driver'] = 'pwned';
    $config['password_minimum_score'] = 3;
    $config['password_algorithm'] = 'sha512-crypt';
    $config['password_httpapi_url'] = 'https://veximurl.example.com/userchangepassword.php';
    $config['password_httpapi_method'] = 'POST';
    $config['password_httpapi_var_user'] = 'user';
    $config['password_httpapi_var_curpass'] = 'curpass';
    $config['password_httpapi_var_newpass'] = 'newpass';
```

### DIFF
--- a/vexim/userchangepassword.php
+++ b/vexim/userchangepassword.php
@@ -1,0 +1,59 @@
+<?php
+/*
+  Endpoint for external interface like roundcube to change the users password
+*/
+
+include_once dirname(__FILE__) . '/config/variables.php';
+include_once dirname(__FILE__) . '/config/functions.php';
+
+# first check if we have sufficient post variables to achieve a successful login... if not the login fails immediately
+if (!isset($_POST['user']) || $_POST['user']==='' 
+    || !isset($_POST['curpass']) || $_POST['curpass']===''
+    || !isset($_POST['newpass']) || $_POST['newpass']===''){
+    http_response_code(400);
+    die("invalid arguments");
+}
+
+# sql statement based on username
+$query = "SELECT u.user_id,u.username,u.crypt
+FROM users as u JOIN domains as d ON d.domain_id = u.domain_id
+WHERE u.username=:username
+AND u.enabled = 1
+AND d.enabled = 1
+AND u.type IN ('local', 'piped')";
+$sth = $dbh->prepare($query);
+$success = $sth->execute(array(':username'=>$_POST['user']));
+if(!$success) {
+  print_r($sth->errorInfo());
+  die();
+}
+if ($sth->rowCount()!=1) {
+  http_response_code(403);
+  die('invalid user or password');
+}
+
+$row = $sth->fetch();
+$cryptedpass = crypt_password($_POST['curpass'], $row['crypt']);
+
+if ($cryptedpass !== $row['crypt']) {
+   http_response_code(403);
+   die('invalid user or password');
+}
+
+if (!password_strengthcheck($_POST['newpass'])) {
+    http_response_code(400);
+    die("week password");
+}
+
+$cryptedpass = crypt_password($_POST['newpass']);
+
+$query = "UPDATE users SET crypt=:crypt WHERE user_id=:user_id";
+$sth = $dbh->prepare($query);
+$success = $sth->execute(array(':crypt'=>$cryptedpass, ':user_id'=>$row['user_id']));
+if (!$success) {
+  print_r($sth->errorInfo());
+  die();
+}
+
+http_response_code(200);
+exit;
--- a/vexim/api/change-user-password.php
+++ b/vexim/api/change-user-password.php
@@ -6,6 +6,11 @@
 include_once dirname(__FILE__) . '/config/variables.php';
 include_once dirname(__FILE__) . '/config/functions.php';
 
+if(!$AllowUserLogin) {
+    http_response_code(403);
+    die("user login is forbidden");
+}
+
 # first check if we have sufficient post variables to achieve a successful login... if not the login fails immediately
 if (!isset($_POST['user']) || $_POST['user']==='' 
     || !isset($_POST['curpass']) || $_POST['curpass']===''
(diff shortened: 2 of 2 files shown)


vexim/userchangepassword.php
```diff
@@ -0,0 +1,59 @@
...
+# sql statement based on username
+$query = "SELECT u.user_id,u.username,u.crypt
+FROM users as u JOIN domains as d ON d.domain_id = u.domain_id
+WHERE u.username=:username
+AND u.enabled = 1
+AND d.enabled = 1
+AND u.type IN ('local', 'piped')";
+$sth = $dbh->prepare($query);
+$success = $sth->execute(array(':username'=>$_POST['user']));
+if(!$success) {
```
rimas-kudelis, line 26: I don't think we should return a success status code here.

vexim/userchangepassword.php
```diff
@@ -0,0 +1,59 @@
...
+$row = $sth->fetch();
+$cryptedpass = crypt_password($_POST['curpass'], $row['crypt']);
+
+if ($cryptedpass !== $row['crypt']) {
+   http_response_code(403);
+   die('invalid user or password');
+}
+
+if (!password_strengthcheck($_POST['newpass'])) {
+    http_response_code(400);
```
rimas-kudelis, line 44: 422 fits better here.

vexim/userchangepassword.php
```diff
@@ -0,0 +1,59 @@
...
+    die("week password");
+}
+
+$cryptedpass = crypt_password($_POST['newpass']);
+
+$query = "UPDATE users SET crypt=:crypt WHERE user_id=:user_id";
+$sth = $dbh->prepare($query);
+$success = $sth->execute(array(':crypt'=>$cryptedpass, ':user_id'=>$row['user_id']));
+if (!$success) {
```
rimas-kudelis, line 53: Should probably return status code 500.

vexim/userchangepassword.php
```diff
@@ -0,0 +1,59 @@
...
+  print_r($sth->errorInfo());
+  die();
+}
+
+http_response_code(200);
+exit;
```
rimas-kudelis, line 59: These two lines seem unnecessary.

rimas-kudelis (commented):
I'm not fond of adding new features to this extremely aged codebase, but at least this PR doesn't change the database, so I'm not killing it right away.

The whole functionality should at least depend on either `$AllowUserLogin`, or a new variable though. It's not a given that every admin wants to allow their users to change their email passwords as simply as this.

Also, the file could probably be named differently to make it stand out more among files used by the Vexim panel itself.

vexim/userchangepassword.php
Zetanova: This is a code snipped from the original login, the sql query is a bit modded to nerrow by type.

vexim/userchangepassword.php
Zetanova: This is a code snipped from the original login.
It makes an sql error print out and die's standard die() code is 500 ?

Zetanova:
I will add `$AllowUserLogin`  am using vexim now for two decades and seeing it the first time :)

vexim/userchangepassword.php
rimas-kudelis: No, `die()` is an alias for `exit()` and it doesn't affect the status code unless you pass it as an argument. See [exit](https://www.php.net/manual/en/function.exit.php).

vexim/userchangepassword.php
rimas-kudelis: Still:
- 200 is the default response code, unless you override it earlier in the code (which we don't do unless we exit the script right afterwards).
- No other script in Vexim is including this one. However, should somebody write a script which would include this one, there's no reason for us to exit here. Our work is done here, but maybe the including script's isn't.

vexim/userchangepassword.php
rimas-kudelis: That doesn't mean it can't be improved, especially since this script will act as a RESTful API endpoint of sorts.

Zetanova (commented):
applied changes

rimas-kudelis:
> applied changes

Won't you address that last comment as well?

vexim/userchangepassword.php
Zetanova: > * No other script in Vexim is including this one. However, should somebody write a script which would include this one, there's no reason for us to exit here. Our work is done here, but maybe the including script's isn't.

I have exactly the reverse opinion about it, I made the commit and removed the guard.

rimas-kudelis: thanks!

rimas-kudelis: Hm, indeed it does look quite weird now with the error handling being the last block in the file and not even a comment afterwards. How about a compromise in form of a `die('Your password has been changed successfully');` line at the end?

rimas-kudelis: Also, do you mind if I squash your commits into one?

Zetanova: please squash it, 
I still don't know how to do it myself in git,
or better to say a decline to learn it (SVN)

About the issue of future script inclusion and the exit at the end:
- To make it possible, all error handler would need to throw an exception and not use die/exit
- Instead of stopping the execution, all decline path's would need to return a result with an error+code

This alone makes a script inclusion in a snippet pattern not possible. 
The real live scenario would be that the functional code would be extracted into an function
to be useable from the endpoint and a future user. 

I have an docker image (already made 2019) and never made it public,
would have spend somebody else some headache.

I only made the API endpoint because I could not find a simple function for sha512-crypt in nodejs for n8n.io,
it seams this pattern is only supported by php and dovecot

rimas-kudelis: Yes, exceptions are the right way to throw errors. But I'll say it again, this code is very old and is hardly maintained anymore. I have an itch to rewrite it using Symfony framework, and would even like to provide a working Docker Compose configuration to make it easy to run Vexim as a set of containerized services, but none of that has materialized yet, because there's just not enough motivation and even general interest. This project is in deep slumber and it's not high on my priority list.

PR #294, opened 2025-03-09
Update vexim-acl-check-rcpt.conf

Udera:
Nixspam list is discontinued:
https://hostblogger.de/blog/archives/7353-Die-AEra-der-ix.dnsbl.manitu.net-geht-zu-Ende.html

### DIFF
--- a/docs/vexim-acl-check-rcpt.conf
+++ b/docs/vexim-acl-check-rcpt.conf
@@ -67,4 +67,4 @@
 # To use a DNSBL the best would be to set up a caching dns server (e.g. unbound) because there is
 # often a rate limit that you will hit when using public DNS servers.
   deny    message       = DNSBL listed at $dnslist_domain\n$dnslist_text
-          dnslists      = zen.spamhaus.org:ix.dnsbl.manitu.net
+          dnslists      = zen.spamhaus.org


rimas-kudelis:
@Udera, I bet you didn't expect me to merge this so fast! :)

Are there any other good lists we could use? I personally am using bl.spamcop.net in addition to spamhaus, but no matter how much spam I report, I keep receiving pretty much same messages, so I've grown skeptical about spamcop and resorted to maintaining local blacklists of hosts and senders.

runout-at:
Maybe something has changed in the exim config and it is not necessary in `vexim-acl-check-rcpt.conf`.

I have in `/etc/exim4/conf.d/main/00_local_listmacrosdefs`

```
CHECK_RCPT_IP_DNSBLS = zen.spamhaus.org : sbl.spamhaus.org : bl.spamcop.net : cbl.abuseat.org : bl.drmx.org : spam.spamrats.com
```

and in `conf.d/acl/30_exim4-config_check_rcpt`

```
  # Check against classic DNS "black" lists (DNSBLs) which list
  # sender IP addresses
  .ifdef CHECK_RCPT_IP_DNSBLS
  warn
    dnslists = CHECK_RCPT_IP_DNSBLS
    add_header = X-Warning: $sender_host_address is listed at $dnslist_domain ($dnslist_value: $dnslist_text)
    log_message = $sender_host_address is listed at $dnslist_domain ($dnslist_value: $dnslist_text)
  .endif

```

rimas-kudelis:
@runout-at, hmm, this seems like pretty standard Debian/Ubuntu/etc setup (it looks like I also use it similarly myself). However, this config warns only, it doesn't reject emails, while ours does.

Ah snap, maybe that's why it seemed to me like reporting spam to spamcop has no effect! These emails were only tagged with an X-Warning header instead of being rejected. Damn!

runout-at:
Maybe you would not like to drop an email for just a single positive on one BL and so it would be better to define BLs in spamassassin or even use policyd-weight.

Eg I use one file per BL like this `/etc/spamassassin/surriel.cf`:

```
header     RCVD_IN_PSBL   eval:check_rbl('psbl-lastexternal','psbl.surriel.com.')
describe   RCVD_IN_PSBL   Received via a relay in PSBL (psbl.surriel.com)
tflags     RCVD_IN_PSBL   net
score      RCVD_IN_PSBL   1 # please adjust the score value
```

rimas-kudelis:
OMG, it's getting complicated by the minute. :exploding_head:

runout-at:
You could use the X-Warning header from exims CHECK_RCPT_IP_DNSBLS in spamassassin like this:

```
header   L_XWARNING_LISTED         X-Warning =~ /.*is\slisted.*/i
score    L_XWARNING_LISTED         0.5
describe L_XWARNING_LISTED         Listed in DNSBL
```

I'm not sure how high the score should be.

Udera:
I found these two:
rbl.0spam.org:all.s5h.net

I prefer to use lists with no false positives and directly reject mails. For other lists, I'd rather add them to spamassassin and adjust the points there.

PR #296, opened 2026-06-07
Escape user-supplied fields when displaying account details

kaluang:
A mailbox user can freely set their real name, vacation message, forward address and header-blocking filters. These values are printed into the account pages without escaping, so stored markup runs when the page is rendered.

The case that matters is the admin views: a normal user can put e.g.

    <script>...</script>

in their real name or vacation message, and it then executes in the domain admin's session when the admin lists users in `adminuser.php` or opens one in `adminuserchange.php`.

Wrapped the affected fields in the existing `html_escape()` helper across `userchange.php`, `adminuserchange.php` and `adminuser.php`. Values that are already escaped, constant, or charset-restricted (localpart/domain) are left alone to keep the diff minimal.

### DIFF
--- a/vexim/adminuser.php
+++ b/vexim/adminuser.php
@@ -96,28 +96,28 @@
             . $row['localpart']
             . '">';
           print '<img class="trash" title="Delete '
-            . $row['realname']
+            . html_escape($row['realname'])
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
Thank you for both PRs @kaluang!

PR #297, opened 2026-06-08
Fix domain capability checks on the user edit form

kaluang:
The checks in adminuserchangesubmit.php that decide whether on_piped,
on_avscan and on_spamassassin may be enabled use a single `=`, so they
assign 1 to $row[...] and the condition is always true. The add form
(adminuseraddsubmit.php) uses `==` and behaves as intended.

on_piped is the main issue: a piped user's smtp value is used directly
as the pipe_transport command in the virtual_domains router, so while
the check is broken a domain admin can switch a user to piped delivery
even when the domain doesn't have pipe enabled. on_avscan and
on_spamassassin are less of an issue because the routers also check the
domain flag, but they have the same typo.

Changes the three comparisons to ==.

### DIFF
--- a/vexim/adminuserchangesubmit.php
+++ b/vexim/adminuserchangesubmit.php
@@ -84,18 +84,18 @@
     }
   }
   # Do some checking, to make sure the user is ALLOWED to make these changes
-  if ((isset($_POST['on_piped'])) && ($row['pipe'] = 1)) {
+  if ((isset($_POST['on_piped'])) && ($row['pipe'] == 1)) {
     $_POST['on_piped'] = 1;
   } else {
     $_POST['on_piped'] = 0;
   }
-  if ((isset($_POST['on_avscan'])) && ($row['avscan'] = 1)) {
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
Hi, thanks for this PR!

Not sure I'm a fan of this intended approach at all though: 
- first of all, I think we should check for `on_piped` in Exim too. If we don't, it's a bug
- secondly, overwriting submitted values seems a bit wrong. IMO, we should probably just ignore them and either hide the fields in question, or render them disabled in the UI (I haven't checked, maybe we do that already).

Anyway, current code is wrong, so I'm merging this. Thanks again!

kaluang:
Thanks, and agreed on both!

I opened a new PR #300 for the Exim side: a virtual_piped router that only pipes when on_piped and the domain's pipe flag are both set, plus forbid_pipe on virtual_domains. The forms already hide those checkboxes when the domain lacks the capability, so this just adds the server-side guarantee. (Left the catchall/group routers as-is since their data isn't the raw smtp field. I would suggest putting this change somewhere into the Readme to notify existing admins that they should implement this manually on existing installations

PR #298, opened 2026-06-08
Stop vacation autoreplies from looping on automated mail

kaluang:
The vacation autoresponder replies to anything that isn't flagged with Precedence: junk/bulk/list. It doesn't look at Auto-Submitted, and the autoreply itself doesn't set that header, so two accounts that are both on vacation keep replying to each other, and replies also go out to other automated senders.

Set Auto-Submitted: auto-replied on the autoreply, and skip incoming mail that already carries an Auto-Submitted header. That breaks the account-to-account loop and avoids answering automated mail, while still replying to normal messages.

This doesn't add once/once_repeat rate limiting, which needs a writable state file and is specific to each install.

### DIFF
--- a/docs/debian-conf.d/router/250_vexim_virtual_domains
+++ b/docs/debian-conf.d/router/250_vexim_virtual_domains
@@ -6,6 +6,7 @@ virtual_vacation:
   driver = accept
   domains = +local_domains
   condition = ${if and { {!match {$h_precedence:}{(?i)junk|bulk|list}} \
+                         {!match {$h_auto-submitted:}{(?i)auto-}} \
                          {eq {${lookup mysql{select users.on_vacation from users,domains \
                                 where localpart = '${quote_mysql:$local_part}' \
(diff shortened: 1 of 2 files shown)


rimas-kudelis:
@kaluang, could you point to a relevant RFC which talks about the `Auto-Submitted` header here, for the sake of completeness?

Edit: nevermind. It's https://www.rfc-editor.org/info/rfc3834/#section-5.

kaluang:
@rimas-kudelis Sure:
RFC 3834 (Recommendations for Automatic Responses to Electronic Mail):
§3.1.7 — an automatic response SHOULD carry Auto-Submitted: auto-replied; that's the header the transport now adds.
§2 — a responder SHOULD NOT reply to a message that already carries an Auto-Submitted header with any value other than no; 
that's the new router condition. The (?i)auto- match covers auto-replied / auto-generated / auto-notified (the IANA keywords), i.e. everything except no.

PR #302, opened 2026-06-08
Scope the block-filter delete to the logged-in user

kaluang:
userblocksubmit.php deletes a blocklists row by block_id with no ownership check, so any logged-in user could delete another user's header-blocking filter by passing an arbitrary block_id. The insert just below already scopes to the session user, and adminuserblocksubmit.php scopes its delete too.

Add user_id and domain_id from the session to the DELETE's WHERE clause.

### DIFF
--- a/vexim/userblocksubmit.php
+++ b/vexim/userblocksubmit.php
@@ -5,9 +5,9 @@
   include_once dirname(__FILE__) . "/config/httpheaders.php";
 
   if ($_GET['action'] == "delete") {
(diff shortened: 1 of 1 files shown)


rimas-kudelis:
@kaluang, what's the point of checking `domain_id`? user IDs are unique, aren't they?

kaluang:
you are right, user_id is the unique primary key.  I had copied it from adminuserblocksubmit.php but there it is neccessary, here it is not.  i have updated the PR

PR #303, opened 2026-06-08
Store a locked password for aliases that don't log in

kaluang:
When an alias is added without a password the code generated a random one, but these accounts only forward mail and aren't meant to log in. Store '*' instead — the conventional "locked account" marker. No crypt() output is ever '*', so the password check fails uniformly for every consumer of the users table (Exim crypteq, Dovecot CRYPT, Courier authmysql, and the web login), without relying on how each treats a NULL/empty crypt field.

(Replaces the earlier approach of generating a random password; see discussion below.)

### DIFF
--- a/vexim/adminaliasaddsubmit.php
+++ b/vexim/adminaliasaddsubmit.php
@@ -30,12 +30,9 @@
   } else {
     $_POST['on_spamassassin'] = 0;
   }
-  # If a password wasn't specified, create a randomised 128bit password
-  if (($_POST['clear'] === "") && ($_POST['vclear'] === "")) {
-    $junk = md5(rand().time().rand());
-    $_POST['clear'] = $junk;
-    $_POST['vclear'] = $junk;
-  }
+  # Aliases that only forward don't log in. When no password is given, store '*'
+  # (a locked account that no crypt() output can match) instead of generating one.
+  $nologin = ($_POST['clear'] === "") && ($_POST['vclear'] === "");
 
   # aliases must have a localpart defined
   if ($_POST['localpart']==''){
@@ -70,8 +67,8 @@
     }
   }
   $aliasto = implode(",",$forwardto);
-  if (validate_password($_POST['clear'], $_POST['vclear'])) {
-    if (!password_strengthcheck($_POST['clear'])) {
+  if ($nologin || validate_password($_POST['clear'], $_POST['vclear'])) {
+    if (!$nologin && !password_strengthcheck($_POST['clear'])) {
       header ("Location: adminalias.php?weakpass={$_POST['localpart']}");
       die;
     }
@@ -87,7 +84,7 @@
        ':localpart' => $_POST['localpart'],
        ':username' => $_POST['localpart'] . '@' . $_SESSION['domain'],
        ':domain_id' => $_SESSION['domain_id'],
-       ':crypt' => crypt_password($_POST['clear']),
+       ':crypt' => $nologin ? '*' : crypt_password($_POST['clear']),
        ':smtp' => $aliasto,
        ':pop' => $aliasto,
        ':realname' => $_POST['realname'],


rimas-kudelis:
I wonder why we even set a random password for accounts which shouldn't log in. `crypt` field is nullable, so I think we should be able to just set it to null instead, thus ensuring that no password works. What do you think?

runout-at:
I don't see a proper way to indicate that an account is not allowed to log in. Did I miss something?

If setting the password to 'NULL', how does other software handle this? Can problems occur on with other software? This DB is not only used by exim. Think about dovecot, courier, webmail clients,...

Actually, if nothing speaks definitely against using 'NULL' I think its a good idea.

kaluang:
Exim and the web login will automatically fail when the password is set to NULL, Dovecot would fail on default but can be set to nopassword and then would allow a login, Courier (which i also use) has some versions that might accept a NULL password, that is therefore not a great solution. We could set the crypt to '*', that would always fail a login on any setup.

rimas-kudelis:
Let's not forget we're not talking about an empty password here, we're talking about a NULL hash value. Did those versions of Courier treat a NULL hash as always valid?

+1 for using `*` as a hash in that case.

rimas-kudelis:
So Courier does treat NULL hashes as valid?

kaluang:
> Did those versions of Courier treat a NULL hash as always valid?

I did a quick search and it appears so. the '*' appears to be generally more resilient, in case vexim is used in another 20 years ;)

rimas-kudelis:
@runout-at, do you mind the asterisk as hash? If not, we can merge this in.

kaluang:
> So Courier does treat NULL hashes as valid?

I checked again, keep in mind this is just a quick research: I could not find a courier-authlib version that would acutally accept the NULL as valid, So it is probably not an actual threat right now, but the way the way the authentication is built in authmysql there is a chance that this introduces a risk.
The dovecot issue is real though, when nopassword is set the server will still require a password if one is set, if the crypt field is empty or NULL it will allow access without a password. this might be a desired behavior but I can not see why someone might want that.

runout-at:
I'm ok with an asterisk too.

runout-at:
Should something be asked/reported at courier/dovecot/...? If they have a policy to handle such values?

Or does a more general policy exist? Like in a RFC.

rimas-kudelis:
Regarding Dovecot, I guess it makes sense. If you have an option to let users in without a password, and that option only lets users in when they actually have no password, it makes sense that you would allow signalling that they have no password by setting their password hash to `null` in addition to using a hash of an empty string.

runout-at:
I'm not sure if this is intended by dovecot or just a flaw. In the IT-world of today I just see possible security issues arising enabling such behavior.

rimas-kudelis:
https://doc.dovecot.org/2.4.4/core/config/auth/passdb.html#nopassword

This is actually documented, but here's the thing: at least to my understanding, `nopassword` must be returned as an extra field from the specific lookup. We never return such field from our lookups, so, in my opinion, the null password should work as intended.

And if we assume that we are responsible for unsupported modifications, then `'*'` could as well pose a problem, should the user decide to use cleartext passswords.

@kaluang, this is for your consideration. :)

kaluang:
I'm mean, I'm fine either way, generally it would be good to add an option to the form to disable user logins. Using the existing get_random_bytes() would probably also be fine for that

runout-at:
Anyway we would do good to throw this case somewhere in our documentation. We are aware of this and if for specific deployments there is a modification the admin of this is responsible to do it in a save way.

rimas-kudelis:
Do we use null values then or keep the asterisk?

runout-at:
I just added a page 'security' to the wiki. Feel free to enhance/change it.

https://github.com/vexim/vexim2/wiki/Security

PR #304, opened 2026-06-09
Add CSRF protection to the web interface

kaluang:
Add a per-session CSRF token that must accompany every state-changing request.

config/csrf.php holds the whole mechanism: csrf_token() creates/returns the token, csrf_input() renders the hidden form field, and csrf_verify() checks it. It depends only on core PHP so httpheaders.php can include it early.

httpheaders.php verifies the token on every POST in one place, so all the POST handlers are covered without changing them. The delete actions that use GET (user/alias/group delete and the block-filter delete) call csrf_verify() directly and carry the token in their link/form. Every form includes the token with csrf_input().

The API endpoint (api/change-user-password.php) is intentionally not covered: it has no session, authenticates with the current password, and is called by external clients.

Heads-up on overlap with #301: both change httpheaders.php right after session_start(), so whichever of the two is merged second will need a small rebase there. There's no behavioural conflict — session_regenerate_id() in #301 keeps the session token.

### DIFF
--- a/vexim/config/csrf.php
+++ b/vexim/config/csrf.php
@@ -1,0 +1,40 @@
+<?php
+  /*
+    Cross-site request forgery protection.
+
+    A random token is stored in the session and must accompany every
+    state-changing request. httpheaders.php verifies it on every POST;
+    the few delete actions that use GET call csrf_verify() directly.
+    Forms include the token with csrf_input().
+
+    This file has no dependencies beyond core PHP so it can be included
+    early (from httpheaders.php) without worrying about include order.
+  */
+
+  // Return the session's CSRF token, creating it on first use.
+  function csrf_token()
+  {
+    if (empty($_SESSION['csrf_token'])) {
+      $_SESSION['csrf_token'] = bin2hex(random_bytes(32));
+    }
+    return $_SESSION['csrf_token'];
+  }
+
+  // A hidden field carrying the token. Place inside every state-changing form.
+  function csrf_input()
+  {
+    // The token is hex (bin2hex), so it needs no HTML escaping.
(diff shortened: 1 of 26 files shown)


vexim/config/csrf.php
```diff
@@ -0,0 +1,40 @@
...
+      $_SESSION['csrf_token'] = bin2hex(random_bytes(32));
+    }
+    return $_SESSION['csrf_token'];
+  }
+
+  // A hidden field carrying the token. Place inside every state-changing form.
+  function csrf_input()
+  {
+    return '<input type="hidden" name="csrf_token" value="'
+      . htmlspecialchars(csrf_token(), ENT_QUOTES, 'UTF-8') . '">';
```
rimas-kudelis: Isn't the token guaranteed to be `[0-9a-f]` per logic above? If so, no escaping is needed, is it?

rimas-kudelis:
Other than the nit above, all seems good, although I'm a bit surprised that explicit call to `csrf_verify()` appears much more often than one to `csrf_token()`.

vexim/config/csrf.php
kaluang: you are right, there is no escaping needed, i will correct that.

kaluang:
the csrf_token() is almost always called indirectly through csrf_input(), which every form uses, so a token is generated on every form render. The only two direct csrf_token() calls are the block-delete links, where the token goes into a URL instead of a hidden field.
csrf_verify() shows up more because it's the single central call for all POST in httpheaders.php, plus one call in each of the five GET delete handlers.
I did also rebase the PR

rimas-kudelis:
That's what surprised me: there are two block-delete links, but five GET handlers. I see, we have some forms with method="get". Jeez! :facepalm:

PR #307, opened 2026-06-11
Added LMTP transport to Dovecot and Exim configs

MrSleeps:
LMTP is a more modern way of delivering mail, these are the configs I use on my system to make that magic work.

### DIFF
--- a/docs/clients/dovecot/etc/dovecot/conf.d/auth-vexim-sql.conf.ext
+++ b/docs/clients/dovecot/etc/dovecot/conf.d/auth-vexim-sql.conf.ext
@@ -1,0 +1,44 @@
+# File: /etc/dovecot/conf.d/auth-vexim-sql.conf.ext
+# Authentication for Vexim SQL users
+
+# Database driver and connection
+sql_driver = mysql
+
+mysql localhost {
+    dbname = vexim
+    user = vexim
+    password = YOUR_PASSWORD
+}
+
+# Password database
+passdb sql {
+    default_password_scheme = BCRYPT
+
+    query = \
+        SELECT username AS user, \
+               CONCAT('{CRYPT}', crypt) AS password \
+        FROM users \
+        WHERE username = '%{original_user}' \
+          AND enabled = 1 \
+          AND type IN ('local', 'piped')
+}
+
+# User database
+userdb sql {
+    query = \
+        SELECT pop AS home, \
+               uid, \
+               gid, \
+               CONCAT('maildir:', smtp) AS mail, \
+               CONCAT('*:bytes=', quota, 'M') AS quota_rule \
+        FROM users \
+        WHERE username = '%{original_user}' \
+          AND enabled = 1 \
+          AND type IN ('local', 'piped')
+
+    iterate_query = \
+        SELECT username AS user \
+        FROM users \
+        WHERE enabled = 1 \
+          AND type IN ('local', 'piped')
+}
--- a/docs/clients/dovecot/etc/dovecot/conf.d/10-auth.conf
+++ b/docs/clients/dovecot/etc/dovecot/conf.d/10-auth.conf
@@ -1,0 +1,121 @@
+##
+## Authentication processes
+##
+
+# Enable LOGIN command and all other plaintext authentications even if
+# SSL/TLS is not used (LOGINDISABLED capability). Note that if the remote IP
+# matches the local IP (ie. you're connecting from the same computer), the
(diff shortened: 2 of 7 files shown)


runout-at:
There is already a PR for this feature: #247 (running on my live system since 2017)

This PR #247 splits out the 'pipe' router. If we want to do it in this or another way is currently in discussion in PR #300

runout-at:
I had a short check if you do something different.

- Renaming of the main vexim router is not a very good idea. 250_vexim_virtual_domains -> 250_vexim_virtual_domains_lmtp
- My version uses some macros. This way settings can be adapted without changing the router and transport. (.../main/00_vexim_listmacrosdefs)
- You do provide some dovecot config in whole files which would substitute the packaged config. It would be better not to touch the original config if possible. This causes headaches during configuration on merging with other specific changes and later problems on upgrades which wants to overwrite your changes. I do this by throwing this config changes in files like: 11-master-override.conf, 21-lmtp-override.conf, ...

But I like the idea to add suggestions for configuring dovecot. Possibly this could go to the wiki page too: https://github.com/vexim/vexim2/wiki/Server-configuration:-Dovecot-v2.x

runout-at:
your router is not backwards compatible.

I provide it this way:

```
  .ifdef LMTP_COMMAND LMTP_SOCKET
    file_transport = virtual_delivery_lmtp
  .else
    file_transport = virtual_delivery
  .endif
```

runout-at:
Another question because I'm not sure we need the 'piped' user in the dovecot sql query. It's on my server too. But do we need it?

A 'pipe' account would give the mail to a command and not to a mailbox. So dovecot won't find the file structure.

Should 'piped' be removed?

```
user_query =
...
    AND type IN ('local', 'piped')
...
```

MrSleeps:
> I had a short check if you do something different.
> 
> * Renaming of the main vexim router is not a very good idea. 250_vexim_virtual_domains -> 250_vexim_virtual_domains_lmtp
> * My version uses some macros. This way settings can be adapted without changing the router and transport. (.../main/00_vexim_listmacrosdefs)
> * You do provide some dovecot config in whole files which would substitute the packaged config. It would be better not to touch the original config if possible. This causes headaches during configuration on merging with other specific changes and later problems on upgrades which wants to overwrite your changes. I do this by throwing this config changes in files like: 11-master-override.conf, 21-lmtp-override.conf, ...
> 
> But I like the idea to add suggestions for configuring dovecot. Possibly this could go to the wiki page too: https://github.com/vexim/vexim2/wiki/Server-configuration:-Dovecot-v2.x

Makes sense (to me) to have 2 different example files so that way the user can choose which way they want to install it.

Dovecot, was just an example. Been running 2.4 since it came out and there's been a fair few upgrades that have never overwritten it.. It also gives a example of what it will look like from a fresh install. Feel free to mod it.

MrSleeps:
> 
> Should 'piped' be removed?
> 


In my opinion, yes. I disabled it about 10 years ago, allowing users to run unchecked scripts that are triggered  via  an email is, imo, a gaping hole in a systems security.

runout-at:
OK. I have removed the 'piped' user from the queries on the wiki page.

https://github.com/vexim/vexim2/wiki/Server-configuration:-Dovecot-v2.x/_compare/9be7d8432054ffce3ca5c9b6a6701ddc1486bd1f...a253d6ca13221dc24b1e4402efb9fc2f1b3808d6

docs/clients/dovecot/etc/dovecot/conf.d/auth-vexim-sql.conf.ext
```diff
@@ -0,0 +1,44 @@
...
+passdb sql {
+    default_password_scheme = BCRYPT
+
+    query = \
+        SELECT username AS user, \
+               CONCAT('{CRYPT}', crypt) AS password \
+        FROM users \
+        WHERE username = '%{original_user}' \
+          AND enabled = 1 \
+          AND type IN ('local', 'piped')
```
runout-at, line 23: You still accept the 'piped' user here and in the 2 sql statements below.

MrSleeps:
VExim2 supports pipe.. We have 0 ways of having an email delivered to a pipe command in our systems but it's supported by VExim. Our views on pipe don't match all other sys-admins views. 

If Vexim is moving away from supporting pipe then it needs to go but as it stands pipes  are supported.

But that was taken from our internal mail server, so that is pure laziness in updating the config ;)

Issue #97, opened 2015-11-04
DKIM


MrSleeps:
Is there any reason why DKIM hasn't been touched so far? I'm pondering converting my text based dkim lookups into SQL lookups to integrate with vexim but before I dive in... :)

rimas-kudelis:
We're not against DKIM, if that's what you're asking. :)

By the way, did you really just overwrote all the work you did while porting Vexim to Bootstrap?..

Udera:
DKIM is a feature I'd like to implement as well. I don't know if it is worth spending too much time on old code when we are considering a new rewrite: https://groups.google.com/forum/#!topic/vexim/3c-EWu7EgE8

Recently, we implemented some security fixes and didn't really add new features. Now, we struggle to push a new version to the benefit of all users.

MrSleeps:
Yeah I did delete everything, I figured I'd start from scratch, wasn't happy with the results. So I'm currently rewriting it.

I'll have a play with DKIM over the next few days, I've got a basic version working but still relies on having a text file so I need to figure out what I'm doing wrong on there, then I'll push it to here

Udera:
Do you want to put the keys into the database?

Udera:
I needed to implement DKIM for a few domains. I did a quick&dirty solution:

- a textfile with all domains that should be signed (-> could be handled by database)
- all domains use the same key
- all domains use the same selector

We could add a button for the siteadmin to turn DKIM on and off for a specific domain (and for the domain admin). That would be really easy.

Managing the certificates via the interface gives the domain-admin more options. Should we be able to generate certificates via webinterface? If we import certificates, we need to check the syntax. The problem is that the domain-owner does not get any feedback from the mailserver if anything fails. In the end, is there a real advantage for the admin?

MrSleeps:
Sorry, GitHub never notified me that you'd replied!

I did roughly the same hack as you, free time has been a rarity, so one day I will sit down and work through a proper solution. Though, that hack does work and has served me well for the last 10 months or so.. I can't see the problem with domains using the same key etc. I don't know what other peoples opinions are..

My personal preference is to allow domain admins to do it via the webpage (I'm slowly untangling my own admin interface (it's currently all linked in with other system adminy stuff)) which I'll upload at some point in the (hopefully not so) distant future. If it's set up the same way as we have both done, it shouldn't be too difficult to implement?

rimas-kudelis:
@Udera:
[...]

PR #308, opened 2026-06-11
Storing DKIM information in the database

MrSleeps:
DKIM keys stored in the vexim database, been running this for years (been a couple of sql updates over the years). Forgot to update the issue I posted almost 10 years ago about it.

Left dkim_sign_headers up to Exims standard signing because despite many hours of testing, could never get a stable dkim result with it.. Exims built in has worked fine for it (it's a busy mailserver).

The SQL is basic but leaves in the option to add sign_headers if you so want.

Cf #97

### DIFF
--- a/docs/debian-conf.d/main/00_vexim_listmacrosdefs
+++ b/docs/debian-conf.d/main/00_vexim_listmacrosdefs
@@ -37,6 +37,19 @@ keep_environment =
 add_environment = MAIN_ADD_ENVIRONMENT
 .endif
 
+# DKIM
+# Requires dkim.sql migration to be run
+DKIM_DOMAIN = ${lc:${domain:$h_from:}}
+DKIM_SELECTOR = ${lookup mysql {SELECT d.selector FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}fail}
+DKIM_PRIVATE_KEY = ${lookup mysql {SELECT d.private_key FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}fail}
+DKIM_CANON = relaxed
+DKIM_STRICT = 0
+DKIM_TIMESTAMPS = 0
+
+# Disable DKIM verification for incoming mail (since we don't want to verify inbound)
+DKIM_VERIFY_DOMAIN = ${if match_domain{$sender_address_domain}{+local_domains}{0}{1}}
+
+
 # validation of sending mailserver
 #CHECK_RCPT_REVERSE_DNS = true
 #CHECK_RCPT_SPF = true
--- a/docs/dkim/dkim.sql
+++ b/docs/dkim/dkim.sql
@@ -1,0 +1,15 @@
+CREATE TABLE `dkim` (
+    `dkim_id`       INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
+    `domain_id`     INT(10) UNSIGNED NOT NULL,
+    `selector`      VARCHAR(63) NOT NULL DEFAULT 'default',
+    `private_key`   TEXT NOT NULL,
+    `public_key`    TEXT NOT NULL,
+    `canonical`     ENUM('simple', 'relaxed') NOT NULL DEFAULT 'relaxed',
+    `sign_headers`  VARCHAR(255) DEFAULT NULL,  -- Optional: custom headers to sign
+    `enabled`       TINYINT(1) NOT NULL DEFAULT 1,
+    `created_at`    TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    `updated_at`    TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    PRIMARY KEY (`dkim_id`),
+    UNIQUE KEY `domain_selector` (`domain_id`, `selector`),
+    FOREIGN KEY (`domain_id`) REFERENCES `domains` (`domain_id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;


docs/debian-conf.d/main/00_vexim_listmacrosdefs
```diff
@@ -37,6 +37,19 @@ keep_environment =
...
 .endif
 
+# DKIM
+# Requires dkim.sql migration to be run
+DKIM_DOMAIN = ${lc:${domain:$h_from:}}
+DKIM_SELECTOR = ${lookup mysql {SELECT d.selector FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}fail}
+DKIM_PRIVATE_KEY = ${lookup mysql {SELECT d.private_key FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}fail}
+DKIM_CANON = relaxed
+DKIM_STRICT = 0
+DKIM_TIMESTAMPS = 0
```
rimas-kudelis, line 47: Why are these hardcoded, especially `DKIM_CANON` which has its own column in the database?

Also, I don't like how you didn't define `DKIM_SIGN_HEADERS` on purpose, but kept the database column. IMO, a commented out macro with an explanation about why it's been commented out would fit here better.

docs/debian-conf.d/main/00_vexim_listmacrosdefs
```diff
@@ -37,6 +37,19 @@ keep_environment =
...
+
+# Disable DKIM verification for incoming mail (since we don't want to verify inbound)
+DKIM_VERIFY_DOMAIN = ${if match_domain{$sender_address_domain}{+local_domains}{0}{1}}
```
rimas-kudelis, line 50: Is this a new macro? I don't see it being used on Bookworm.

docs/dkim/dkim.sql
rimas-kudelis, line 1: If we ship with this functionality, then this file should go to `docs/setup/migrations`, and if not, maybe all of this could just go to the Wiki?

rimas-kudelis (commented):
In addition to stuff below: 
1. since this doesn't come with any UI, maybe it should wait until we get rid of the default admin panel?
2. The table structure doesn't seem to protect against having duplicate records for the same domain, or having none. What happens in these cases?

docs/debian-conf.d/main/00_vexim_listmacrosdefs
rimas-kudelis: By the way, here's my `DKIM_SIGN_HEADERS`:

> DKIM_SIGN_HEADERS = +Content-Type:+MIME-Version:+Date:+Message-ID:+To:+Subject:+From:+Sender:=Reply-To:+Cc:+Content-ID:+Content-Description:=Resent-Date:=Resent-From:=Resent-Sender:=Resent-To:=Resent-Cc:=Resent-Message-ID:+In-Reply-To:+References:=List-Id:=List-Help:=List-Unsubscribe:=List-Subscribe:=List-Post:=List-Owner:=List-Archive

I think it's been working nicely for me for quite a while now.

runout-at: Is there a reason to not use the default which conforms to RFC?

rimas-kudelis: I don't really remember what problem I was solving by editing these values myself. It's probably what I said [here](https://github.com/vexim/vexim2/issues/97#issuecomment-570900027).

MrSleeps: Mainly it's as it is to start a discussion on what is needed/wanted. 

We hard code them at work because we have a lot of email accounts and reducing sql look ups is never a bad thing (you can set the canon via sql if you really wanted, hence it's there).

Same with DKIM sign headers, those are left to Exims defaults but if people want to change them per domain they can?

You have to remember there was no activity on this repo, I had no idea what direction you wanted to go with it. 

> The table structure doesn't seem to protect against having duplicate records for the same domain, or having none. What happens in these cases?

Why would it? How do you rotate secrets if you just delete the old one? Any mail in transit  that hasn't caught up with the new key in the dns would fail?

As for having no key, it would fail. That's a personal preference, in todays world of email not having a dkim key and trying to send email is kind of pointless.

You could do something like

```
DKIM_SELECTOR = ${lookup mysql {SELECT d.selector FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}{}}
DKIM_PRIVATE_KEY = ${lookup mysql {SELECT d.private_key FROM dkim d JOIN domains dom ON d.domain_id = dom.domain_id WHERE dom.domain = '${quote_mysql:DKIM_DOMAIN}' AND d.enabled = 1}{$value}{}}
```

And have in the transports have something that looks something like:

```
dkim_selector = ${if def:DKIM_SELECTOR {DKIM_SELECTOR}{}}
dkim_private_key = ${if def:DKIM_PRIVATE_KEY {DKIM_PRIVATE_KEY}{}}
```

Which wouldn't sign the email if it didn't have a key (in theory, currently untested)

docs/debian-conf.d/main/00_vexim_listmacrosdefs
MrSleeps: Didn't actually mean to leave that in there..

Was working on an acl that uses it

```
acl_smtp_dkim:
  # Only verify external senders
  condition = ${if eq {DKIM_VERIFY_DOMAIN}{1}}
  
  # If verification fails, set a variable
  warn
    dkim_status = fail
    set acl_m_dkim_fail = yes
    log_message = DKIM failure from $sender_address_domain: $dkim_verify_reason

  # Accept otherwise
  accept
```
  
  But exim does (at least on trixie) have this on the check_rcpt acl
  
`  .ifdef DISABLE_DKIM_VERIFY`
  
  So it seemed kinda pointless

docs/dkim/dkim.sql
MrSleeps: I think in 2026 it needs some kind of dkim in the core vexim setup, DKIM has been around since the mid 00s.

docs/debian-conf.d/main/00_vexim_listmacrosdefs
runout-at: How do you rotate the public keys on your DNS servers? Automaded/Manual?

MrSleeps: Fully automated (thousands of domains would take days!), every 3 months a new one gets generated, dns runs off a sql backend (pdns) so it's dead simple. Generate key, update the dns sql and add the new key to email database. A week later the old key gets deleted. The old key is kept so dns propagation isn't an issue.

MrSleeps: Forgot to mention, the keys get a new selector when they are rotated so it would be jan2026 for one and the new one would be jun2026 for example.

PR #316, opened 2026-06-15
Remove pipe users, add lmtp, bugfixes

runout-at:
I couldn't change PR #247 anymore. Thats why I offer a new PR.

This PR would introduce the LMTP transport which is also in PR #307. The difference is that this version let you define in `lismacrodefs` if you would like to use a socket or a command:

```
#LMTP_COMMAND = /some/local/lmtp/delivery/program
LMTP_SOCKET = /var/run/dovecot/lmtp
```
Fallback is a normal delivery if no LMTP SOCKET/COMMAND is defined.

This PR removes pipe user functionality as discussed in PR #300 from the exim router. It seems we don't have the corresponding transport in our files.

I found and fixed a small bug which occurs in some special cases when using addresses with VEXIM_LOCALPART_SUFFIX.

In a query in the virtual_dom_groups router i fixed: use the email address instead of the username. As in veximpy the username might not be the address.

In `listmacrodefs` I added

```
CHECK_MIME_LOCAL_ACL_FILE = /etc/exim4/vexim-acl-check-mime.conf
```
(used in acl/50_exim4-config_check_mime - at least in my debian setup)

### DIFF
--- a/docs/debian-conf.d/main/00_vexim_listmacrosdefs
+++ b/docs/debian-conf.d/main/00_vexim_listmacrosdefs
@@ -90,7 +90,7 @@ i# lmtp delivery to an lmtp capable application like dovecot
 # this will enable the LMTP routers and set a command or socket in the LMTP transport
 # ! Set only one of the macros !
 #LMTP_COMMAND = /some/local/lmtp/delivery/program
-LMTP_SOCKET = /var/run/dovecot/lmtp
+#LMTP_SOCKET = /var/run/dovecot/lmtp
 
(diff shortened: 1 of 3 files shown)


rimas-kudelis:
Do the three parts in the subject really have to go together in the same PR?

runout-at:
> Do the three parts in the subject really have to go together in the same PR?

not really. but it would be lots of work to separate it. I did just copy my running router config.

runout-at:
I did think about splitting this PR but the change for LMTP would need to split out the pipe router from virtual_delivery anyways and and the review work would be more. 

The bugfixes are really small.

PR #317, opened 2026-06-16
change is_public in groups table to tinyint

rimas-kudelis:
Replaces #257


runout-at (approved):
Thank you very much! Looks good to me!

rimas-kudelis:
Do we have anyone to test with PostgreSQL?

rimas-kudelis:
Oh I forgot. We will also need a migration SQL file for this. It should first alter the data, then change column type.

Until we decide to make a release, I guess it could be called descriptively by what it does. Before tagging new version, we'd merge all these changes into one file.

rimas-kudelis:
Alright, so we need a migration here.

